Issue #1, opened 2017-05-20
Add Remote Logging for bots


itsthejoker:
Currently we have a key set up for Sentry, but it only allows one team member to access it. This is a problem going forward, so the next contender worth looking at is bugsnag.

TheLonelyGhost:
If we go the self-hosted route, we could also look at [Errbit](https://errbit.com/docs/master/) with the [Airbrake client for Python](https://pypi.python.org/pypi/airbrake)

itsthejoker:
Right now we need to avoid the self hosted route since both the bot and the logging would be on the same network. That way we'd get no warning if the internet went out, which has already happened once since I've moved.

Once we're on DigitalOcean or somewhere else, this is much less of an issue.

itsthejoker:
Current branch: https://github.com/itsthejoker/TranscribersOfReddit/tree/the-logging-update

PR #19, opened 2017-05-21
#1 - Adding bugsnag logging tags and configuration

itsthejoker:
Ticket: #1 

## Description:

Adds Bugsnag integration into the bots. The API key is stored in an environment variable on the local machine, so we don't have to worry about errant calls getting sent during testing.


itsthejoker:
This was merged into #20 and merged into master.

Issue #17, opened 2017-05-21
Disable username mentions by OP of post


itsthejoker:
We do not want the bot to act as a free service for OPs, so the bot should ignore username calls by OPs of media and only respond if a different account summons it.

PR #21, opened 2017-05-21
#17 - Ignore username mentions by OPs

arfie:
Fixes [issue #17](https://waffle.io/itsthejoker/TranscribersOfReddit/cards/5920e366a1a2b700261ba314)

### DIFF
--- a/tor/core/mentions.py
+++ b/tor/core/mentions.py
@@ -44,8 +44,7 @@ def process_mention(mention, r, tor, redis_server, config):
         parent.title = '"' + parent.title + '"'
 
     # Ignore requests made by the OP of content or the OP of the submission
-    if mention.author.id == parent.author.id or \
-            mention.author.id == submission.author.id:
+    if mention.author == parent.author or mention.author == submission.author:
         logging.info(
             'Ignoring mention by OP u/{} on ID {}'.format(mention.author,
                                                           mention.parent_id)
--- a/tor/core/mentions.py
+++ b/tor/core/mentions.py
@@ -25,6 +25,7 @@ def process_mention(mention, r, tor, redis_server, config):
     :return: None.
     """
 
+    # The submission the current mention was made in
     submission = r.submission(id=clean_id(mention.link_id))
 
     # We have to do this entire parent / parent_permalink thing twice because
@@ -44,7 +45,7 @@ def process_mention(mention, r, tor, redis_server, config):
         parent.title = '"' + parent.title + '"'
 
     # Ignore requests made by the OP of content or the OP of the submission
-    if mention.author == parent.author or mention.author == submission.author:
+    if mention.author == submission.author:
         logging.info(
             'Ignoring mention by OP u/{} on ID {}'.format(mention.author,
                                                           mention.parent_id)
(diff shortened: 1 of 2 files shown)


tor/core/mentions.py
```diff
@@ -40,6 +41,14 @@ def process_mention(mention, r, tor, redis_server, config):
         # format that sucker so it looks right in the template.
         parent.title = '"' + parent.title + '"'
 
+    # Ignore requests made by the OP of content
+    if mention.author.id == parent.author.id:
```
itsthejoker: What happens if they hide it by replying to another comment? So:

```
This is the post that's probably not OC
  I'm a top level comment!
    Wait, I can't read it
      oh yeah, /u/transcribersofreddit
```

arfie:
Right, now it will block mentions when either it's the poster of the parent comment, or the OP (of the submission). People would still be able to perform requests from within their comment itself, or (if they're not OP) from a deeper subcomment. I could look into fixing the latter tomorrow; I don't see a good solution to the first problem but I think we can live with it.

tor/core/mentions.py
```diff
@@ -24,6 +24,9 @@ def process_mention(mention, r, tor, redis_server, config):
     :param redis_server: Active redis instance.
     :return: None.
     """
+
+    submission = r.submission(id=clean_id(mention.parent_id))
```
itsthejoker: So fun fact: if you call r.submission (instead of r.comment) on a comment, it will error when you try to interact with it, which is why we keep it after that if / else -- that way we know it's good to go.

tor/core/mentions.py
```diff
@@ -35,11 +38,20 @@ def process_mention(mention, r, tor, redis_server, config):
...
         # this is a post.
-        parent = r.submission(id=clean_id(mention.parent_id))
+        parent = submission
         parent_permalink = parent.permalink
         # format that sucker so it looks right in the template.
         parent.title = '"' + parent.title + '"'
 
+    # Ignore requests made by the OP of content or the OP of the submission
+    if mention.author.id == parent.author.id or \
+            mention.author.id == submission.author.id:
```
itsthejoker: I think we're definitely on the right track here. Could this be simplified to just `if mention.author == submission.author`? Also, there's another function that may be of use here. I have a helper function, located in tor.helpers.reddit_ids called get_parent_post_id -- it's expensive, but it's entire goal in life is to take any comment object and return the top submission ID. Is that something that might be able to clean this up a little?

arfie: I found a property `link_id` in `Comment` on [here](https://github.com/reddit/reddit/wiki/JSON) which seems to do the exact same thing (well, I obviously can't test the bot but from playing with PRAW in a shell it seems to work, and [this](https://www.reddit.com/r/redditdev/comments/29ewkl/praw_get_submissionid_of_a_comment/cikcydx/) says the same). I'll use that for now, seems a bit strange to me that nested comments wouldn't have a direct reference to the submission they're in.

Comparing author properties seems to work as well yes, they are different instances but seem to overload the equality operator.

arfie:
Actually, I see now that PRAW has built in methods `submission()` and `parent()` in `Comment` which do the same thing but look cleaner. I'm not going to change to those now of course, but it might be worth looking into later.

PR #31, opened 2017-06-19
Initial refactor

TheLonelyGhost:
I know this is a huge set of changes, but mainly they surround some basic concepts:

- Pip packaging
- Trailing spaces (and other flake8 linting)
- Removing imports and other non-essential functionality from `__init__.py`
- Automated testing
- Documentation clarity

Any and all feedback is welcome. If necessary I can split these up into multiple PR's too.

**EDIT: Code review changes**

- [x] Add in Cython again to the dependencies
- [x] Changelog entries
- [x] Add suggested libs to docs (`better_exceptions`, `hiredis`)
- [x] Add better documentation of thoughts moving forward for refactoring (e.g., `tor.core.config`)


requirements.txt
```diff
@@ -1,11 +1,7 @@
 praw==4.4.0
 redis<3.0.0
-hiredis
```
itsthejoker: Perhaps I haven't gotten there yet, but even though we don't call this library, it's used as a speedup by the redis library.

requirements.txt
```diff
@@ -1,11 +1,7 @@
...
-better_exceptions
```
itsthejoker: Even if we don't use it in production, having this available for debugging is incredibly useful. Is there a reason we want to nuke it?

requirements.txt
```diff
@@ -1,11 +1,7 @@
...
 addict
-cython
```
itsthejoker: Cython is required to compile tesserocr, and tersserocr doesn't always pull it in on its own.

test/core/test_config.py
```diff
@@ -0,0 +1,51 @@
...
+
+        assert type(SITE_CONFIG.bs_api_key) is str or \
+            SITE_CONFIG.bs_api_key is None
+
+        assert type(SITE_CONFIG.slack_api_url) is str or \
+            SITE_CONFIG.slack_api_url is None
+
+
+if __name__ == '__main__':
+    unittest.main()
```
itsthejoker: This is a great start. Easy to move from this into the more complex testing with PRAW.

tor/core/config.py
```diff
@@ -0,0 +1,128 @@
...
+    import bugsnag
+except ImportError:
+    bugsnag = None
+
+from tor import __version__
+
+
+class BaseConfig:
+    '''
+    '''
```
itsthejoker: So just to make sure I understand, these are deliberately blank because they get populated by `initialize()`, yah?

requirements.txt
TheLonelyGhost: I didn't notice it imported anywhere. Figured it was a dependency of a dependency or something.

requirements.txt
TheLonelyGhost: Nuking it from the dependencies doesn't remove it from being installed if don manually. How are you using it? Is that on the dev side or when debugging in production?

requirements.txt
itsthejoker: https://github.com/redis/hiredis-py It's primarily for pipeline'd redis commands, which we don't actually use (although we might need to as we move more towards the daemonized side of things). I'm open to removing it until we actually need it.

requirements.txt
TheLonelyGhost: Fair point. I actually read up on Cython and how to use it. Didn't see it used in this project so I figured it was a nested dependency.

I suppose it should be documented why this is being manually added. I'll hop on that and readd this dependency.

requirements.txt
itsthejoker: I used it for both, but now that we have bugsnag it's not really necessary in production. We can add it to the list of "useful things to have from a dev standpoint". It cleans up tracebacks and tells you what's in every variable at the time of the explosion.

tor/__init__.py
```diff
@@ -1,49 +1 @@
-import bugsnag
-from addict import Dict
-
 __version__ = '2.6.14'
```
itsthejoker: This is definitely worthy of at LEAST 2.6 😛

test/core/test_config.py
TheLonelyGhost: I was also looking into how python does mocks. Looks promising for use with PRAW. I'd much prefer to pull the raw PRAW calls into an actor object. That encases the entire action, then mock that out for other tests.

itsthejoker: Oh yeah, Mock (especially MagicMock) is stupidly powerful. It's only useful when you know exactly what the response should be though, so implementing it might take a little.

bin/run
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh
```
itsthejoker: Would it make more sense to replace this with a makefile? `make run`, `make test`, all that?

tor/core/config.py
TheLonelyGhost: Essentially. It's more along the lines of establishing that all child classes will have a certain interface. It's a kind of throwback from strongly typed, compiled languages which rely on interfaces. Not too useful right now, but it will be with future refactoring.

itsthejoker: Sounds good. More clarity is basically always a good thing.

tor/__init__.py
TheLonelyGhost: Heh, I was avoiding touching a version increment until after merging a PR. It's especially useful if there are multiple PRs in a single release. Didn't want to presume.

itsthejoker: We don't have a good system for incrementing versions. My system loosely follows semantic versioning, but only when I remember to update the version number. Then I usually bump it one or two extra to make up for the releases I missed. It's very unofficial.

bin/run
TheLonelyGhost: I would, but the convention is to use `python setup.py` as a task runner. Try `python setup.py - -
help-commands` from the project root sometime.

As for using a shell script instead of a Makefile, I wanted to defer all unknown commands to `python setup.py`. I also know how foreign the Makefile syntax is to some people and figured this would be more community-friendly.

itsthejoker: Ooh, okay. That makes sense. I have no experience with anything that's _not_ makefiles, so I'll give this a shot :)

tor/__init__.py
TheLonelyGhost: Maybe we should follow guidelines at http://keepachangelog.com here? It enforces a process for release without meaning to, making semver much easier.

itsthejoker: Just gave it a read through; I'm fine with that implementing that. Seems like a sane way to do it :)

test/core/test_config.py
TheLonelyGhost: If we care about the response. For a lot of tests, we might only care about making sure it doesn't call out to the internet. Maybe making sure it is called at all, but not about the response. At least not every time.

I guess it might not be called a "mock" in all of those cases, but somehow overriding the method's internal workings to not be called is the idea. I've never had too big of an issue with testing internal workings when I've been able to mock/stub/fake methods out and refactor the original code.

itsthejoker:
@arfie do you have any thoughts on this before we merge?

TheLonelyGhost:
I'd like to fix some of the nits brought up before merging, but I'm always open for more review

itsthejoker:
Well, yes; blame my sleep-addled brain for that one. It's been edited.

TheLonelyGhost:
Added the libraries we're removing (until they're needed) to a newly created Github wiki: https://github.com/itsthejoker/TranscribersOfReddit/wiki

TheLonelyGhost:
Alright, @itsthejoker, I think I'm done fixing/adding things. Wanna review the new changes? @arfie, do you have any input?

tor/__init__.py
TheLonelyGhost: Added the CHANGELOG.md file and CONTRIBUTING.md in a subsequent commit, per the recommendations in http://keepachangelog.com. 👍

PR #34, opened 2017-07-01
Archivebot basics

arfie:


tor/archiver.py
```diff
@@ -0,0 +1,82 @@
...
+    # TODO we can use .submissions(end=unixtime) apparently
+
+    for post in tor.new():
+        # [META] - do nothing
+        # [UNCLAIMED] - remove
+        # [COMPLETED] - remove and x-post to r/tor_archive
+        # [IN PROGRESS] - do nothing (should discuss)
+        flair = post.link_flair_css_class
+
+        if flair not in ('unclaimed', 'completed'):
```
itsthejoker: We have a dot-notation-accessible dict for this -- https://github.com/TranscribersOfReddit/TranscribersOfReddit/blob/master/tor/helpers/flair.py#L12

arfie: Right, but I'm looking for the correct key rather than looking up the value in the flair dict. So dot notation is not going to help much :)

itsthejoker:
```
post_flair = post.link_flair_css_class

if post_flair != flair.unclaimed or post_flair != flair.completed:
```

Isn't this what you're trying to do?

arfie: I'm comparing CSS class rather than flair text, so there's no need to look it up in the flair dict.

itsthejoker: Ah, okay. In that case, the "completed" flair class is actually `transcriptioncomplete`

tor/archiver.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if flair not in ('unclaimed', 'transcriptioncomplete'):
+            continue
+
+        # find the original post subreddit
+        # take it in lowercase so the config is case insensitive
+        post_subreddit = subreddit_from_url(post.url).lower()
+
+        # hours until a post from this subreddit should be archived
+        hours = config.archive_time_subreddits.get(
```
itsthejoker: Great use of .get()!

tor/archiver.py
```diff
@@ -0,0 +1,82 @@
...
+    r = Reddit('bot_archiver')
+
+    configure_logging(config)
+    logging.basicConfig(filename='archiver.log')
+
+    tor = configure_tor(r, config)
+
+    archive = r.subreddit('ToR_Archive')
+
+    try:
```
itsthejoker: You'll probably also want to catch soft crashes from Reddit not communicating:
```
except (
        prawcore.exceptions.RequestException,
        prawcore.exceptions.ServerError,
        prawcore.exceptions.Forbidden
    ) as e:
        logging.warning(
            '{} - Issue communicating with Reddit. Sleeping for 60s!'
            ''.format(e)
        )
        time.sleep(60)
```

tor/archiver.py
```diff
@@ -0,0 +1,94 @@
...
+        while True:
+            run(tor, config, archive)
+            sleep(300)  # 5 minutes
+
+    except KeyboardInterrupt:
+        logging.info('Received keyboard interrupt! Shutting down!')
+        sys.exit(0)
+
+    except (
+            prawcore.exceptions.RequestException,
```
itsthejoker: nitpick 1/2: 4 spaces :)

tor/archiver.py
```diff
@@ -0,0 +1,94 @@
+import sys
+import logging
+
+import prawcore
+from praw import Reddit
+
+from tor import config
+from tor.core.initialize import configure_logging
+from tor.core.initialize import configure_tor
+from tor.helpers.misc import explode_gracefully, subreddit_from_url
```
itsthejoker: nitpick 2/2: separate lines, unless we do it elsewhere in the codebase.

tor/archiver.py
arfie: whoops, didn't paste right

tor/archiver.py
TheLonelyGhost: I've seen it done with parentheses like this in several codebases and it seems to follow your requirement. Does this version bother you?

```
from foo import (
   Fizz,
   Buzz,
   LoremIpsum,
)
```

itsthejoker: That's a valid way of doing it I suppose; I just have a thing against putting multiple imports on one line because when you have a shitload of imports (like some of the projects I have at work) it's difficult to scan the import list and immediately see where an import is.

I'm not terribly preferential to either way of doing it (each on a separate line or what you've got there), just that we decide on it and it goes into the style guide.

PR #44, opened 2017-09-05
truncate titles that are too long

arfie:


arfie:
I'm quite sure the maximum is 21, based on this: https://github.com/reddit/reddit/blob/master/r2/r2/models/subreddit.py#L114

PR #45, opened 2017-09-06
Validation Update

itsthejoker:
Since inception, we've been running a "soft" version of our validation which only checked to make sure that the person claiming had made a top level comment on the thread in question. This was supposed to be modified a few months ago and I just completely forgot.

This change verifies that the user in question has made a top level comment and verifies that the key is in said comment before signing off on it.

This PR also includes a fluff change of allowing randomly generated responses if anyone says "thanks" to the bot, because it has happened a few times now.


tor/core/validation.py
```diff
@@ -73,11 +73,16 @@ def verified_posted_transcript(post, config):
...
     linked_resource = config.r.submission(top_parent.id_from_url(top_parent.url))
     # get rid of the "See More Comments" crap
     linked_resource.comments.replace_more(limit=0)
     for top_level_comment in linked_resource.comments.list():
-        if post.author == top_level_comment.author:
-            return True
+        if _author_check(post, top_level_comment):
+            # we can do an 'and' statement above, but that would require
+            # running the header check on every comment as well as the author
+            # check and we want to save cycles wherever we can.
```
arfie: That's not true, is it? When using `and`, the right side is only even evaluated if the left side is true. So using `and` would be the same as having nested `if`s.

tor/strings/responses.py
```diff
@@ -52,3 +54,19 @@
...
+    'https://media.giphy.com/media/ClcWrARkrq1GM/giphy.gif',
+    'http://i.imgur.com/QMPRHW9.gif',
+    'https://media.giphy.com/media/j5QcmXoFWl4Q0/giphy.gif',
+    'http://www.reactiongifs.com/r/a-b-yay.gif',
+    'https://media.tenor.com/images/7dcd26f4d9a2eaa246cd4ea9967d2336/tenor.gif',
+    'http://media0.giphy.com/media/wTzPSxZBqSove/giphy.gif'
+]
+
+youre_welcome = (  # just format that sucker as a link and send it back
+    '[Hey, you\'re welcome!]({})'.format(random.choice(thumbs_up_gifs))
```
arfie: Not catastrophical, but this would only evaluate once and use the same gif every instance of the bot that is run.

tor/core/validation.py
itsthejoker: screenshot 2017-09-06 09 54 08

You're totally right. Oops.

tor/strings/responses.py
itsthejoker: I distinctly remember thinking "I wonder if that wouldn't work" and then dismissing it. Great catch!

TheLonelyGhost: Yep, it would definitely insert one image consistently until the bot is completely restarted. Possibly python's caching mechanisms (`*.pyc`) might mess with this too, not sure.

PR #46, opened 2017-09-06
Updates changelog to include missed releases

TheLonelyGhost:
Looks like we haven't been too diligent on updating the changelog for each release. Now all version changes (except one flub with `2.6.14` -> `2.6.1` -> `2.7.0`) are [tagged in git](https://github.com/TranscribersOfReddit/TranscribersOfReddit/tags).

This PR is a followup by attempting to describe the changes between `v2.7.1` -> `v3.0.0` and between `v3.0.0` -> `v3.0.1`

Also prunes stale `console_scripts` in `setup.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Nothing yet...
 
 - Adds Travis CI support, enforcing support for python 3.6
 - Updates docs for `pip install` using a git url
-- Adds CLI tools `tor-moderator`, `tor-apprentice`, and `tor-archivist` to PATH (instead of `python path/to/bot.py`)
+- Adds CLI tool `tor-moderator` to PATH (instead of `python tor/main.py`)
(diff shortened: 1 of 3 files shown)


NGTmeaty:
@TheLonelyGhost could we also update the readme in this pull?

TheLonelyGhost:
NO! ^_^

... sure, gimme a sec

TheLonelyGhost:
@NGTmeaty Is that what you wanted? What else do you see in the README that needs to be updated?

itsthejoker:
I think he was referring more specifically to the information about the other bots which is no longer valid.

TheLonelyGhost:
You know, I forgot we followed through with splitting up the bots into separate repositories. I removed the entries in `setup.py` so it's less confusing and updated associated docs (changelog included).

Is that what we're talking about with updating the readme?

NGTmeaty:
looks good :+1:

PR #47, opened 2017-09-20
Bugfix: be able to handle Reddit messages

itsthejoker:
Very rarely, we may receive messages from Reddit itself. In this case, these messages will not have an author, which kind of breaks things. This PR allows us to catch those messages, ping slack, and continue on with our day.


tor/core/inbox.py
```diff
@@ -31,6 +32,16 @@ def check_inbox(config):
     replies = []
     # grab all of our messages and filter
     for item in config.r.inbox.unread(limit=None):
+        # Very rarely we may actually get a message from Reddit itself.
+        # In this case, there will be no author attribute.
+        if item.author is None:
+            send_to_slack(
+                'We received a message without an author. Subject: {}'
+                ''.format(item.subject), config
```
TheLonelyGhost: Wait, does this work with python? I thought `format()` would act on the empty string, then concatenate the two strings that are (relatively) side-by-side.

itsthejoker: screenshot 2017-09-20 14 57 10

TheLonelyGhost: Cool! I figured you knew what you were doing there. The python strings-next-to-each-other-are-concatenated concept breaks my brain.

PR #49, opened 2017-09-30
Slack CoC Notification update

itsthejoker:
Allows the CoC notifications we receive to have a clickable link in them now. This change has been okay'd as usable by Sam.

screen shot 2017-09-30 at 10 39 13 am

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -50,8 +50,8 @@ def process_coc(post, config):
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_slack(
-            '<http://www.reddit.com/u/{}|u/{}> has just accepted the CoC!'.format(
-                post.author.name, post.author.name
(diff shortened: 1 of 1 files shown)


tor/core/user_interaction.py
```diff
@@ -50,8 +50,8 @@ def process_coc(post, config):
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_slack(
-            'u/{} has just accepted the CoC!'.format(
-                post.author.name
+            '<http://www.reddit.com/u/{}|u/{}> has just accepted the CoC!'.format(
+                post.author.name, post.author.name
```
TheLonelyGhost: We could provide a named argument in the string ({} => {user}) and only include the name once (as a keyword argument) in the format() arguments.

itsthejoker: That... did not occur to me. Yeah, that seems like a decent change.

PR #53, opened 2017-10-08
Add ability to bypass content filters

itsthejoker:
This is a three-part update and will need to be handled with care to make sure all three parts are verified before deploying.

* u/ToR PR: https://github.com/TranscribersOfReddit/TranscribersOfReddit/pull/53
* tor-core PR: https://github.com/TranscribersOfReddit/tor_core/pull/8
* wiki update: https://www.reddit.com/r/TranscribersOfReddit/wiki/format/other and https://www.reddit.com/r/TranscribersOfReddit/wiki/subreddits/domain-filter-bypass

What this PR does is enable adding subreddits to the `domain-filter-bypass` wiki page, allowing all content posted there to be crossposted to r/ToR. This option will currently be used for two subreddits: r/DescriptionPlease and r/Futurology.


tor/core/posts.py
```diff
@@ -103,10 +104,9 @@ def process_post(new_post, config):
         content_type = 'video'
         content_format = config.video_formatting
     else:
-        # how could we get here without fulfilling one of the above
-        # criteria? Just remember: the users will find a way.
+        # This means we pulled from a subreddit's bypassing the filters.
         content_type = 'Unknown'
```
arfie: Maybe it would be better to change that to "Other" in this case?

tor/core/posts.py
```diff
@@ -35,7 +35,8 @@ def check_submissions(subreddit, config):
         if (
             post.domain in config.image_domains or
             post.domain in config.audio_domains or
-            post.domain in config.video_domains
+            post.domain in config.video_domains or
+            subreddit in config.subreddits_domain_filter_bypass
```
TheLonelyGhost: since we're touching this anyway... What about:

```python
if (post.domain in config.image_domains + config.audio_domains + config.video_domains or
    subreddit in config.subreddits_domain_filter_bypass):
```

tor/core/posts.py
```diff
@@ -103,10 +104,9 @@ def process_post(new_post, config):
         content_type = 'video'
         content_format = config.video_formatting
     else:
-        # how could we get here without fulfilling one of the above
-        # criteria? Just remember: the users will find a way.
-        content_type = 'Unknown'
-        content_format = 'Formatting? I think something went wrong here...'
+        # This means we pulled from a subreddit bypassing the filters.
+        content_type = 'Other'
+        content_format = config.other_formatting
```
TheLonelyGhost: Is the `config.other_formatting` value being set somewhere? It looks like it requires 1 of 2 things still:

- update to `tor_core` repo for additional value
- update to initialization of this bot in particular in order to set this value

tor/core/posts.py
itsthejoker: That would work, but is there a performance reason for it or just clarity? I'd probably just format it the same way as above, so I'm not sure overall clarity would change.

tor/core/posts.py
itsthejoker: Both of which are done over in https://github.com/TranscribersOfReddit/tor_core/pull/8/files

TheLonelyGhost: Oops! You're right. It was even put in the description of this PR. 😅

tor/core/posts.py
TheLonelyGhost: Both clarity and a performance reason, as far as I can see. It's negligible as far as the gains of each, but to me it lends to an abstraction of a method like `is_whitelisted_domain()`.

I suppose all of this conditional would be put in there in that case, though. Carry on!

tor/core/posts.py
itsthejoker: nodding

TheLonelyGhost:
@itsthejoker Did you update the changelog?

itsthejoker:
Mmmmno. I should do that.

PR #55, opened 2017-10-16
Partner PR for 3.1.0: Invert Inbox

itsthejoker:
Right now, the inbox pulls in the default PRAW method, which is newest to oldest. If the bot is offline for some amount of time and responses are allowed to build up, that means that "done" comments will be processed before "claim" comments.

This change pulls all the unread messages, puts them in a list, inverts the list, and then processes from oldest to newest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ means we will never make a backwards-incompatible change within a major version 
 Nothing yet...
 
 ## v3.1.0 (2017-10-14)
+- Now processing inbox messages in the correct order
 - Adds support for pulling all submissions from specific subreddits
 
 ## v3.0.3 (2017-10-01)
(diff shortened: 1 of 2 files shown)


TheLonelyGhost:
Popped in here to confirm that there is no way to reverse the sort order of the inbox at the API level. At least not according to the documentation.

Only way to make this more efficient would result in a significant hike in code complexity, as far as I can tell. 👍  on the concepts. Will give a more thorough review later.

arfie:
Is it really that inefficient, though? It still runs in linear time, and doesn't make any additional API calls by reversing the order.

TheLonelyGhost:
@arfie I'm talking less about API calls and more about how much work the bot has to do.

In this case, I was investigating if we could offload the ordering to the server, such that pagination is oldest to newest. I suppose it's negligible if we're keeping up-to-date with the unread messages in the inbox anyway, or we just pay the penalty in order to catch up with all of them. Meh. Not a big enough problem to worry about, you're right.

PR #57, opened 2017-10-17
Guard every reply against DELETED_COMMENT

TheLonelyGhost:
Pretty self-explanatory. Any time we try to do a `post.reply()`, we should be guarding for APIException of type `DELETED_COMMENT` like this. In most cases it should be safe to just drop it.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -1,6 +1,5 @@
 import logging
 import random
-import sys
 
 import praw
 # noinspection PyProtectedMember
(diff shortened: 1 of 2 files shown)


itsthejoker:
Do we still need this block? https://github.com/TranscribersOfReddit/TranscribersOfReddit/blob/7b15cffa9d83ac4dbf1ecae94a820d184271e13c/tor/core/user_interaction.py#L152

tor/core/user_interaction.py
```diff
@@ -74,39 +73,45 @@ def process_claim(post, config):
...
-                top_parent.fullname, post.author
+        # can't claim something that's already claimed
+        elif top_parent.link_flair_text == flair.in_progress:
+            post.reply(_(already_claimed))
+        elif top_parent.link_flair_text == flair.completed:
+            post.reply(_(claim_already_complete))
+
+    except praw.exceptions.APIException as e:
+        if e.error_type == 'DELETED_COMMENT':
+            logging.debug(
```
itsthejoker: Let's check to see if we have a top_parent object and then force-reset the flair back to unclaimed in case it crashed after awarding flair. `flair_post(top_parent, flair.unclaimed)`

tor/core/user_interaction.py
```diff
@@ -74,39 +73,45 @@ def process_claim(post, config):
...
+                'Back up for grabs!'.format(top_parent.fullname)
             )
-        )
-    # can't claim something that's already claimed
-    elif top_parent.link_flair_text == flair.in_progress:
-        post.reply(_(already_claimed))
-    elif top_parent.link_flair_text == flair.completed:
-        post.reply(_(claim_already_complete))
+            return
+        raise  # Re-raise exception if not
```
itsthejoker: Does `raise` in this context actually reraise the exception with all its information? Seems like we would want to raise `e`.

TheLonelyGhost: I thought the same, but:

> When raising (or re-raising) an exception in an except or finally clause __context__ is automatically set to the last exception caught; if the new exception is not handled the traceback that is eventually displayed will include the originating exception(s) and the final exception.

(from [python docs on exceptions](https://docs.python.org/3/library/exceptions.html))

itsthejoker: Huh, I sit corrected. Nice!

tor/core/user_interaction.py
TheLonelyGhost: I interpreted the code being slightly different already. The diff doesn't make this too clear, but it looked like events happen in this order:

1. Check for error states (e.g., no code of conduct for user) and exit with error message if any
2. Fix error state of no status in flair. Set status to unclaimed.
3. Reply to post saying it is now claimed
4. Mark OP with flair saying it's claimed

(1) failing means it didn't modify the flair anyway, since it only checks for error states, replies with an error message, and quits out

(2) failing means it didn't correct an already failing state.

(3) failing is the main error for which we're trying to account. If it fails, (4) won't be executed anyway and flair isn't modified in a way that needs to be reverted.

(4) failing means it has a bot-replied claimed comment, but isn't actually claimed. This should never happen since the OP being deleted means either the bot (original author) or a subreddiit mod or a super mod removed it. In any event, the post is gone and, claimed or not, it cannot be completed anyway.

It keeps going for possible events, but they're also error handling by posting an error message in a reply. No modification of flair that needs to be fixed if it cuts out early.

tor/core/user_interaction.py
TheLonelyGhost: In other languages, calling `raise e` like that would clobber the stack trace while `raise` wouldn't. In python it doesn't appear to clobber in the same way like I originally thought. In any event, I left a comment to clarify what's happening here. :smile:

tor/core/user_interaction.py
TheLonelyGhost: It looks like we modify the flair after replying to the comment already. If the reply fails, which is what throws this `DELETED_COMMENT` error, we don't even execute the flair modification code. Do we really need to reset it then?

itsthejoker: It appears you're right. Nevermind! :D

TheLonelyGhost:
@itsthejoker In response to [your comment earlier](https://github.com/TranscribersOfReddit/TranscribersOfReddit/pull/57#issuecomment-337096720) about the try-except block, yes we still need it.

Per the [PRAW docs](https://praw.readthedocs.io/en/latest/code_overview/exceptions.html#module-praw.exceptions):

> \[PRAW\] includes two main exceptions: `APIException` for when something goes wrong on the server side, and `ClientException` when something goes wrong on the client side. Both of these classes extend `PRAWException`.

They have a common parent, but they are very different classes both semantically and in practice. What I do find odd is that sometimes the error when acting on a deleted comment is raised as a server-side error whereas other times it is raised as a client-side error.

itsthejoker:
Excellent. I'll merge and test this evening.

PR #58, opened 2017-10-19
Prunes unneeded dependencies and fixes (some) tests

TheLonelyGhost:
Dependencies that existed from the OCR bot aren't needed here.

Also fixed some of the tests and marked one as expected to fail, because it isn't passing for some yet unknown reason.


README.md
```diff
@@ -32,13 +32,13 @@ Find the [latest release](https://github.com/TranscribersOfReddit/TranscribersOf
...
 $ git checkout v3.0.1
-$ pip install .
+$ pip install --process-dependency-links .
 ```
 
 OR
 
 ```
-$ pip install 'git+https://github.com/TranscribersOfReddit/TranscribersOfReddit.git@v3.0.1#egg=tor'
+$ pip install --process-dependency-links 'git+https://github.com/TranscribersOfReddit/TranscribersOfReddit.git@v3.0.1#egg=tor'
```
itsthejoker: Let's set this to not use a version number -- at least for now, we can reasonably assume that we want to use the latest version of tor_core.

TheLonelyGhost: Of tor_core? That depends on the setup.py since this has it as a dependency. From context I assume that was a typo and you meant tor. Will change. :+1:

itsthejoker: Yeah, that. o7

PR #62, opened 2017-10-20
Consistent docs and project structure

TheLonelyGhost:
This is a multi-project set of PRs:

- TranscribersOfReddit/ToR_Flair#3
- TranscribersOfReddit/ToR_Archivist#2
- TranscribersOfReddit/tor_core#10
- TranscribersOfReddit/ToR_OCR#8
- TranscribersOfReddit/TranscribersOfReddit#62

All of these pull requests center around the same concept: updating documentation and tooling to be a unified front.

- waffle.io badges were used inconsistently
- some references to the repository A when repository B was the intended target
- messy dependency tracking
- license files varying between `LICENSE` and `LICENSE.txt`, referencing a copyright holder as `TranscribersOfReddit`, which is not a legal entity.
- ... and so on

This merges all of that into a set of clean instructions, descriptive... well, descriptions, and making as much use of setup.py as possible (instead of the split standard with requirements.txt).


TheLonelyGhost:
Merging due to no activity after 11 days. It's also holding up the migration to cloud hosting.

PR #64, opened 2017-10-25
Make the admin commands less buggy slightly

perryprog:
10/10 commit message


arfie:
Looks great! Don't forget to update the changelog.

tor/core/inbox.py
```diff
@@ -57,15 +57,14 @@ def check_inbox(config):
         if 'reload' in item.subject.lower():
             item.mark_read()
             reload_config(item, config)
-            item.reply(
-                'Config reloaded!'
-            )
```
TheLonelyGhost: Yeah, we probably should be replying that it was reloaded if it wasn't (e.g., non-authorized user DM's the command)

tor/core/inbox.py
```diff
@@ -57,15 +57,14 @@ def check_inbox(config):
...
+
             continue
         if 'update' in item.subject.lower():
             item.mark_read()
             update_and_restart(item, config)
             # there's no reason to do anything else here because the process
             # will terminate and respawn
+            # ok so that doesn't actually happen so ignore that
```
TheLonelyGhost: Not sure the point of this comment. Honestly, since the update and restart item isn't working anyway, we should probably just remove this section entirely, but that's an issue for another time.

I might suggest here that we change the wording of the original comment rather than tack on more comments at the end refuting prior ones.

itsthejoker: I'm with @TheLonelyGhost on this one -- since it's not part of the PR, let's remove the comment and leave for later.

TheLonelyGhost:
(Also ignore the status checks, we're still experimenting with the services and configurations)

itsthejoker:
Along with what @arfie said about updating the changelog, please increment tor.__init__.py by one bugfix release (`3.1.0` --> `3.1.1`)

perryprog:
Got it, writing commit right now.

perryprog:
done

perryprog:
also about ci tools and all of that, I would recommend Travis, code beat, and codacy. I don't really like codeclimate too much. (no reason why)

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ means we will never make a backwards-incompatible change within a major version
 - Guard every comment reply in case it gets deleted
 - Prunes unneeded dependencies from before tor_core extraction
 
+## v3.1.1 (2017-10-15)
```
arfie: Should be 2017-10-25, actually

perryprog: Freaking dates

PR #65, opened 2017-11-06
Defers more things to environment variables

TheLonelyGhost:
- Debug mode
- Production mode user reference in praw.ini

### DIFF
--- a/tor/main.py
+++ b/tor/main.py
@@ -1,3 +1,4 @@
+import os
 import time
 
 from tor_core.config import config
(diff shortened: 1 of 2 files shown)


tor/main.py
```diff
@@ -46,9 +47,12 @@ def run(config):
 
 
 def main():
-    build_bot('bot', __version__, full_name='u/ToR')
+    config.debug_mode = bool(os.environ.get('DEBUG_MODE', False))
+    bot_name = 'debug' if config.debug_mode else os.environ.get('BOT_NAME', 'bot')
```
itsthejoker: if all of the bots are running on one machine, do we want to change BOT_NAME to something more specific to u/tor?

TheLonelyGhost: We can set the environment variable name on invocation and it only affects that instance, not the rest of the bots.

PR #66, opened 2017-11-06
Send message to Slack when getting unknown replies, and minor formatting

### LINKED ISSUE #33: Send message to Slack when receiving unprocessable inbox messages
Current process for handling messages is waiting for Reddit to email unread messages to me and then manually following up.

Investigate and implement messaging Slack with the contents of received messages that cannot be processed.

perryprog:
Closes #33 

I also removed two unused imports, and some minor formatting.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -90,31 +90,35 @@ def check_inbox(config):
             if 'i accept' in reply.body.lower():
                 process_coc(reply, config)
                 reply.mark_read()
+                return
 
-            elif 'claim' in reply.body.lower():
+            if 'claim' in reply.body.lower():
                 process_claim(reply, config)
                 reply.mark_read()
+                return
 
-            elif 'done' in reply.body.lower():
+            if 'done' in reply.body.lower():
                 process_done(reply, config)
                 reply.mark_read()
+                return
 
             # trigger on "thanks" and "thank you"
-            elif 'thank' in reply.body.lower():
+            if 'thank' in reply.body.lower():
                 process_thanks(reply, config)
                 reply.mark_read()
+                return
 
-            elif '!override' in reply.body.lower():
+            if '!override' in reply.body.lower():
                 process_override(reply, config)
                 reply.mark_read()
+                return
 
-            elif 'good bot' in reply.body.lower() \
-                    or 'bad bot' in reply.body.lower():
+            if 'good bot' in reply.body.lower() or 'bad bot' in reply.body.lower():
                 # please stop emailing me, I just don't care
                 reply.mark_read()
+                return
 
-            else:
-                send_to_slack('Unknown reply: ' + reply.body)
+            send_to_slack('Unknown reply: ' + reply.body, config)
 
         except (AttributeError, praw.exceptions.ClientException):
             # the only way we should hit this is if somebody comments and then
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -67,7 +67,8 @@ def check_inbox(config):
         if item.subject.lower() == 'ping':
             item.mark_read()
             logging.info(
-                'Received ping from {}. Pong!'.format(item.author.name))
+                'Received ping from {}. Pong!'.format(item.author.name)
+            )
             item.reply('Pong!')
 
     # sort them and create posts where necessary
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -32,7 +32,7 @@ def forward_to_slack(item, config):
 
     logging.info(
         'Received unhandled inbox message from {}. \nSubject: {}\n\nBody: {}'
-        .format(item.author, item.subject, item.body)
+        ''.format(item.author, item.subject, item.body)
     )
 
 
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -28,7 +28,11 @@ def forward_to_slack(item, config):
             author=item.author,
             body=item.body,
             subject=item.subject,
-        ), config
+        ), config)
+
+    logging.info(
+        'Received unhandled inbox message from {}. \nSubject: {}\n\nBody: {}'
+        .format(item.author, item.subject, item.body)
     )
 
 
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -120,8 +120,10 @@ def check_inbox(config):
                 return
 
             send_to_slack(
-                'Unknown reply by {author}: {body}'.format(
-                    author=reply.author, body=reply.body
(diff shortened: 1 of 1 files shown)


perryprog:
Codebeat what's wrong with 7 elif statements??

itsthejoker:
Heh. There's a reason I didn't do elif statements originally, but can't remember why.

TheLonelyGhost:
Generally, I've found they're less flexible. They preclude a mutually exclusive relationship between the different code blocks executed when the condition evaluates to true. Consider the following:

```python
def example_1():
    if A:
        b()
    elif C:
        d()
    e()


def example_2():
    if A:
        b()
    if C:
        d()
    e()


def example_3():
    if A:
        b()
        return
    if C:
        d()
    e()
```

Let's evaluate the 3 example methods above.

- Assume `A` and `C` are falsey. In all 3 examples, only `e()` is called.
- Assume `A` is falsey and `C` is truthy. In all 3 examples, `d()` and `e()` are called.
- Assume `A` and `C` are truthy. `example_1()` calls `b()` and `e()`. `example_2()` calls `b()`, `d()`, and `e()`. `example_3()` calls just `b()`.
- Assume `A` is truthy and `C` is falsey. `example_1()` and `example_2()` each call `b()` and `e()`. `example_3()` only calls `b()`.

Stylistically, I'm more in favor of `example_2()` and `example_3()`'s style because I like to think of it like firing off an event. Multiple events may be fired off and, if you don't want to evaluate any further events, you can just `return` (or `continue` if in a loop) to end it early.

The style in `example_1()` means `b()` and `d()` are mutually exclusive events. They may be, but I typically find more uses where `b()` and `d()` are unrelated streams of events that might fire, only dependent on `A` or `C` respectively.

perryprog:
Did you miss what I just pushed?

perryprog:
And if you want I can put it back to if elif elif else, I didn't see your message when I committed.

tor/core/inbox.py
```diff
@@ -112,9 +112,13 @@ def check_inbox(config):
                 process_override(reply, config)
                 reply.mark_read()
                 return
+
             if 'good bot' in reply.body.lower() or 'bad bot' in reply.body.lower():
                 # please stop emailing me, I just don't care
                 reply.mark_read()
+                return
+
+            send_to_slack('Unknown reply: ' + reply.body, config)
```
TheLonelyGhost: Can we also include the author (if there is one)? Example:

`send_to_slack('Unknown reply by {author}: {body}'.format(author=reply.author, body=reply.body), config)`

perryprog:
Hope you don't mind my formatting 🧐(new emojis cause why not)

TheLonelyGhost:
Also, re: emoji, this is all I see: 
screen shot 2017-11-06 at 21 59 17

tor/core/inbox.py
```diff
@@ -112,9 +113,17 @@ def check_inbox(config):
...
                 reply.mark_read()
                 return
+
             if 'good bot' in reply.body.lower() or 'bad bot' in reply.body.lower():
                 # please stop emailing me, I just don't care
                 reply.mark_read()
+                return
+
+            send_to_slack(
+                'Unknown reply by {author}: {body}'.format(
```
itsthejoker: The only request I have is to also add the subject of the message. Other than that it looks solid.

itsthejoker:
@TheLonelyGhost's PR restructured the inbox a bit, but this should drop in at roughly the same place. You'll just need to resolve the conflicts.

perryprog:
There is basically 0 chance I did that right, I need someone to check that.

tor/core/inbox.py
```diff
@@ -23,6 +22,16 @@
 ]
 
 
+def send_reply_to_slack(item, config):
```
TheLonelyGhost: Do we want to instead rename this method to "forward_to_slack"? It seems more appropriate in the given context.

tor/core/inbox.py
```diff
@@ -23,6 +22,16 @@
...
+    send_to_slack(
+        'Unknown reply by **{author}**, {subject}: {body}'.format(
+            author=item.author,
+            body=item.body,
+            subject=item.subject,
+        ), config
+    )
```
TheLonelyGhost: We might also want to mark this item as read so we don't keep forwarding the message to slack every time the bot goes through the inbox. I might be missing something here, though. Thoughts, @itsthejoker?

tor/core/inbox.py
```diff
@@ -89,6 +98,8 @@ def process_reply(reply, config):
             reply.mark_read()
             return  # Because overrides should stop the world and start fresh
 
+        send_reply_to_slack(reply, config)
```
TheLonelyGhost: According to #33, we're only intending to forward unhandled _inbox messages_ to slack. This seems like it's a handled message, but purposefully ignored since it's a reply to a comment. I could be off-base here. Wanna weigh-in, @itsthejoker?

TheLonelyGhost:
We kind of can't avoid the expletives since that's the entire purpose of some of the regular expressions: To flag potentially abusive comments for moderator review. We might just rework this part to go through automod, but this is still an experiment to see how we like it.

tor/core/inbox.py
TheLonelyGhost: After chatting with him, sounds like he wants _all_ unhandled messages to be sent to slack. This is intended as a replacement for reading email associated with the bot accounts.

Per @itsthejoker, "it needs to be sent to slack, written to disk, and then marked as read"

itsthejoker: and by "written to disk" I mean throw it in a logging Info statement so we can search for it later if we need to.

tor/core/inbox.py
perryprog: Nah, I wanted to spam you guys with messages. /s

tor/core/inbox.py
perryprog: "Written to disk"

What's the format that you want for that?

perryprog:
That's much less exciting :(

tor/core/inbox.py
itsthejoker: Just the same info that's getting sent to slack in case the slack call fails. The slack call, by design, is a "fire and forget" system. There is no check to see if it succeeds -- it just happens. So just like other places we're using it, we want to also put in a logging statement that encapsulates the same information so we have a backup.

`logging.info('Received unhandled inbox message from {0}. \nSubject: {1}\n\nBody: {2}'.format(appropriate_things))

perryprog:
I'm putting the item.mark_read() bit when I can the function, for consistency with the other handlers.

PR #69, opened 2017-11-09
HISTORY VALIDATION!

itsthejoker:
A long time coming -- this should, in theory, allow us to pull transcriptions that were removed by the Reddit global spam filter from the user's history and run the validation against that instead. It's not very many extra steps and it doesn't need to run that often -- it will be extremely useful when volunteers start posting in subs they don't frequent, like r/4chan.

TODO:

- [x] Add additional check that the author of the parent posts match as well as the titles, as this is an avenue for cheating the system
- [x] Update changelog
- [x] Update version number
- [x] Test the sucker

This has been manually tested, but I'll test using the PR before merging after it's approved.


perryprog:
```python
for doohickey in post.author.new(limit=3):
```

QUALITY

itsthejoker:
@perryprog you would not believe how many of my variables start out life as `doohickey`. This one just accidentally made it into the codebase.

TheLonelyGhost:
Yeah, I'm gonna need us to get rid of all the `doohickey`s before we merge this... 😆 

coffee

tor/core/validation.py
```diff
@@ -39,13 +67,17 @@ def verified_posted_transcript(post, config):
...
+            _header_check(top_level_comment, config)
+        ):
             return True
-    # Did it get removed? Check their history.
-    # if _author_history_check:
-    #     return True
-    # return False
+
+    # Did their transcript get flagged by the spam filter? Check their history.
+    return _author_history_check(post, config)
```
TheLonelyGhost: Is there any reason we wouldn't want to solely rely on this history check for the one source of truth for the transcription comment? Looping through the top-level comments seem like extra API calls that aren't needed if the author history would cover it on its own.

tor/core/validation.py
```diff
@@ -15,10 +15,38 @@ def _header_check(reply, config, tor_link=ToR_link):
         return True
 
 
+def _thread_title_check(original_post, history_item):
+    # Verify that the link titles match. On the original post, it will be
+    # removed, but we should still be able to extract the title of the
+    # submission it's on. Then we check to see if the title for that submission
+    # is in the r/ToR post, mirroring the max line truncation that's in
+    # posts.py.
```
TheLonelyGhost: Docstring this, please.

tor/core/validation.py
```diff
@@ -39,13 +67,17 @@ def verified_posted_transcript(post, config):
     # get source link, check all comments, look for root level comment
     # by the author of the post and verify that the key is in their post.
     # Return True if found, False if not.
```
TheLonelyGhost: This seems like a much more concise way of expressing the same concepts as in the docstring above. Maybe this should replace the docstring?

tor/core/validation.py
TheLonelyGhost: Honestly, if a comment is removed, that's the target subreddit's fault not ours. We only care about providing the option of a transcription and giving credit to the author of it. Maybe it means we update the OP on /r/tor with a direct link to the transcription comment. I'm not sure.

itsthejoker: In the greater sense it's the Reddit spam filter that's the cause, but yes, I suppose that's valid. Mostly we've just been counting to see how active the spam filter is / if we need to intervene again with the admins. My worry is that if we don't check for removed comments we'll never know if the spam filter starts removing all of the volunteer posts or not, thus rendering our work useless.

TheLonelyGhost: Ah, so we do want to have some validation tripwire, even if it doesn't catch all of the anti-spam measures. That way we can make sure it has a chance to be useful to others. I'd agree with keeping it as is, in that case.

itsthejoker: Can you elaborate on "validation tripwire"?

TheLonelyGhost: By that I mean it's a tripwire, or early warning system, where we can tell something is either currently, or will very soon go wrong. In this case it's something that won't catch all of the times a transcriber's post is removed, but it will catch some of them and that's better than nothing.

The "validation" part is just pertaining to where the tripwire lives.

tor/core/validation.py
itsthejoker: I feel like the description that's currently in the docstring explains different things. I've moved this bit up there, though.

itsthejoker:
@TheLonelyGhost This is ready for another round of review -- I've got time to do a deploy tomorrow if you have time to take a look at this.

tor/core/validation.py
```diff
@@ -16,34 +26,51 @@ def _header_check(reply, config, tor_link=ToR_link):
...
+    """
     max_title_length = 250
     return (
         history_item.link_title[:max_title_length - 4] in
         original_post.link_title
     )
 
 
+def _thread_author_check(original_post, history_item, config):
+    return (
```
TheLonelyGhost: Not required, but a docstring/comment here to warn that this will hit the API instead of using locally cached data might be nice. This method is short enough that it should be relatively apparent, but I want to make sure we're policing our API usage pretty strictly.

tor/core/validation.py
```diff
@@ -16,34 +26,51 @@ def _header_check(reply, config, tor_link=ToR_link):
...
     :return: True if the post is found in the history, False if not.
     """
     for history_post in post.author.new(limit=5):
         if (
-            _header_check(history_post, config) and
             history_post.is_root and
-            _thread_title_check(post, history_post)
+            _footer_check(history_post, config) and
+            _thread_title_check(post, history_post) and
+            _thread_author_check(post, history_post, config)
```
TheLonelyGhost: I don't recall, does python evaluate these chained conditionals like a binary operator? Something like `a and b and c` where it does not evaluate `c` if `b` is false?

Just trying to separate out ruby and python knowledge. Ruby had some concerns in that area with using `and` versus `&&`.

tor/core/validation.py
```diff
@@ -58,15 +85,17 @@ def verified_posted_transcript(post, config):
     presence of the key. If it's all there, we update their flair and mark
     it complete. Otherwise, we ask them to please contact the mods.
 
+    Process:
+    Get source link, check all comments, look for a root level comment
+    by the author of the post and verify that the key is in their post.
+    Return True if found, False if not.
```
TheLonelyGhost: 💯

tor/core/validation.py
```diff
@@ -75,9 +104,17 @@ def verified_posted_transcript(post, config):
     for top_level_comment in linked_resource.comments.list():
         if (
             _author_check(post, top_level_comment) and
-            _header_check(top_level_comment, config)
+            _footer_check(top_level_comment, config)
```
TheLonelyGhost: I do like how we've renamed the check to, you know, not be a misnomer. `_header_check` should indicate checking the header of the comment, not the footer. 😛 

👍

tor/core/validation.py
```diff
@@ -75,9 +104,17 @@ def verified_posted_transcript(post, config):
...
         ):
             return True
 
     # Did their transcript get flagged by the spam filter? Check their history.
-    return _author_history_check(post, config)
+    if _author_history_check(post, config):
+        send_to_slack(
+            f'Found removed post: {post.submission.shortlink}',
+            config,
+            channel='#removed_posts'
```
TheLonelyGhost: So nice! While slack might not be the authoritative, persistent database of choice, at least we can have some visibility into when posts are removed and not need to dig through the bot's logs. 👍

perryprog:
I'm just commenting here so it looks like I helped

tor/core/validation.py
itsthejoker:
```python
>>> from time import sleep
>>> from timeit import timeit
>>> a = lambda: False
>>> def b():
...  sleep(4)
...  return True
...
>>> def c():
...  sleep(2)
...  return True
...
>>> def test():
...  if a() and b() and c():
...   pass
...
>>> timeit(test, number=1)
4.37488779425621e-06
```

TheLonelyGhost: 😂 I see what you mean, but here's a version that might be clearer for others:

```python
>>> def test(msg, ret):
...     print(msg)
...     return ret
...
>>> if test('first', True) and test('second', False) and test('third', False):
...     print('done')
...
first
second
>>>
```

tor/core/validation.py
itsthejoker: Well, it used to be the header. We just forgot to rename this when that changed.

tor/core/validation.py
itsthejoker: Oh yeah, I'll change that real quick and then merge.

PR #73, opened 2017-11-10
Make all lines fit the required line length, and remove unused imports

perryprog:
@TheLonelyGhost not sure if you think this is necessary, but I want to make sure that we stay consistent.

### DIFF
--- a/tor/helpers/youtube.py
+++ b/tor/helpers/youtube.py
@@ -1,5 +1,4 @@
 import logging
-
 # noinspection PyCompatibility
 from urllib.parse import parse_qs, urlparse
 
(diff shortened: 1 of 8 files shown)


tor/helpers/youtube.py
```diff
@@ -24,8 +23,8 @@ def get_yt_video_id(url):
       Invalid:
         'youtu.be/watch?v=_lOT2p_FCvA',
     """
-    # initial version: http://stackoverflow.com/a/7936523/617185 \
-    #    by Mikhail Kashkin (http://stackoverflow.com/users/85739/mikhail-kashkin)
+    # initial version: http://stackoverflow.com/a/7936523/617185 \ by Mikhail
```
TheLonelyGhost: In this case, I think we can remove the `\` before "by Mikhail". Not entirely certain why it was there to begin with, TBH.

tor/main.py
```diff
@@ -48,7 +48,8 @@ def run(config):
 
 def main():
     config.debug_mode = bool(os.environ.get('DEBUG_MODE', False))
-    bot_name = 'debug' if config.debug_mode else os.environ.get('BOT_NAME', 'bot')
+    bot_name = 'debug' if config.debug_mode \
+        else os.environ.get('BOT_NAME', 'bot')
```
TheLonelyGhost: Since it explodes out to multiple lines (due to line length), why don't we just convert this into a fully-fledged `if`-`else` conditional block here?

perryprog:
Re line length: Actually, I completely agree with you. 
screen shot 2017-11-10 at 3 06 08 pm

And the reason for this PR is simply just consistency.

PR #76, opened 2017-11-19
Use Discord Flair Colors in Subreddit

itsthejoker:
This mirrors the solution we already have in place in Discord to the actual subreddit. We've heard over and over again that people really like this extra thing they can work towards, so this just writes the appropriate css class when their flair is updated. 

The CSS is already in place on the subreddit.

### DIFF
--- a/tor/helpers/flair.py
+++ b/tor/helpers/flair.py
@@ -1,5 +1,7 @@
 import logging
 
+from collections import OrderedDict
+
 from tor_core.helpers import clean_id
 from tor_core.helpers import flair
 
@@ -39,11 +41,31 @@ def update_user_flair(post, config):
     On a successful transcription, this takes the user's current flair,
     increments the counter by one, and stores it back to the subreddit.
 
+    If the user is past 50 transcriptions, select the appropriate flair
+    class and write that back too.
+
     :param post: The post which holds the author information.
     :param config: The global config instance.
     :return: None.
     """
     flair_text = '0 Γ - Beta Tester'
+
+    flair_levels = {
+        0: 'grafeas',
+        51: 'grafeas-green',
+        101: 'grafeas-orange',
+        251: 'grafeas-purple',
+        501: 'grafeas-golden',
+        1000: 'grafeas-diamond'
+    }
+
+    # put them in the right order, since apparently creating the thing as an
+    # OrderedDict doesn't always do that
+    flair_levels = OrderedDict(
+        sorted(
+            flair_levels.items(), key=lambda t: t[0]
+        )
+    )
 
     try:
         # The post object is technically an inbox mention, even though it's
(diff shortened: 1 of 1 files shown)


perryprog:
Oh, I forgot about CSS and stuff

tor/helpers/flair.py
```diff
@@ -61,14 +83,31 @@ def update_user_flair(post, config):
         user_flair = flair_text
 
     if 'Γ' in user_flair:
-        # take their current flair and add one to it
-        new_flair_count = int(user_flair[:user_flair.index('Γ') - 1])
+        # extract their current flair and add one to it
+        new_flair_count = int(user_flair[:user_flair.index('Γ') - 1]) + 1
```
perryprog: Is it possible for someone/thing to have a malformed flair, that this cast could then fail on?

tor/helpers/flair.py
```diff
@@ -39,12 +41,32 @@ def update_user_flair(post, config):
     On a successful transcription, this takes the user's current flair,
```
perryprog: Oh — This function is also way too long and complex, I'd try to break this into several other bits.

tor/helpers/flair.py
itsthejoker: If the flair has the gamma inside it, it's a reasonable assumption that it has been set correctly. We've only had one instance of flair that was "malformed", in which case Reddit for some reason didn't accept the call to actually set it in the first place. If that happens, it gets intercepted in `process_done` and we don't even get this far. Personally I don't think it's an issue -- this particular bit has been running since the very first release without a problem.

tor/helpers/flair.py
itsthejoker: Generally I try to break things up if they meet one of the following criteria:

* over 500 lines long
* has sections that are untestable
* is objectively unreadable without further abstraction

This doesn't hit any of those criteria; why do you think this is too long and complex?

tor/helpers/flair.py
perryprog: In that case it's probably fine

tor/helpers/flair.py
perryprog: 37 lines of code makes Codacy complain :)

It would also reduce the block nesting, which would be good.

itsthejoker: ...bah.

perryprog: my bad, I meant codebeat

tor/helpers/flair.py
```diff
@@ -39,12 +41,32 @@ def update_user_flair(post, config):
...
+        1000: 'grafeas-diamond'
+    }
+
+    # put them in the right order, since apparently creating the thing as an
+    # OrderedDict doesn't always do that
+    flair_levels = OrderedDict(
+        sorted(
+            flair_levels.items(), key=lambda t: t[0]
+        )
+    )
```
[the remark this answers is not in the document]
perryprog: Definitely, that looks much cleaner.

tor/helpers/flair.py
perryprog: ¯\_(ツ)_/¯

tor/helpers/flair.py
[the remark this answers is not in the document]
itsthejoker: I just have a personal vendetta against long `elif` bits. We can definitely go with this, though.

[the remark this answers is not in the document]
perryprog: I like long elif bits....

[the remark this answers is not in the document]
TheLonelyGhost: I'm totally fine with long `elif` bits so long as the method it's in is only dedicated to routing. If there's any logic in there whatsoever, subcontract that out to another method. Makes it easier to test and reduces complexity for comprehension.

PR #79, opened 2017-12-03
Send meta posts to slack

arfie:
This adds a slack message for meta posts, and slightly improves the formatting of the slack messages for unhandled replies.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -22,12 +22,15 @@ MOD_SUPPORT_PHRASES = [
 
 
 def forward_to_slack(item, config):
-    send_to_slack(
-        f'Unhandled inbox reply by *{item.author}* -- *{item.subject}*: '
-        f'{item.body}', config)
+    username = item.author.name
 
+    send_to_slack(
+        f'Unhandled message by '
+        f'<https://reddit.com/user/{username}|u/{username}> -- '
+        f'*{item.subject}*:\n{item.body}', config
+    )
     logging.info(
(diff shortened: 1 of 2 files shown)


tor/core/inbox.py
```diff
@@ -22,12 +22,15 @@
...
 
 def forward_to_slack(item, config):
-    send_to_slack(
-        f'Unhandled inbox reply by *{item.author}* -- *{item.subject}*: '
-        f'{item.body}', config)
+    username = item.author.name
 
+    send_to_slack(
+        f'Unhandled message by '
+        f'<https://reddit.com/user/{username}|u/{username}> -- '
```
perryprog: Why do you have the `{username}|u/{username}`? If it's for visual reasons, can you move the greater than sign to be after the first `{username}`?

tor/helpers/flair.py
```diff
@@ -140,3 +140,6 @@ def set_meta_flair_on_other_posts(config):
                 f'Meta. '
             )
             flair_post(post, flair.meta)
+            send_to_slack(
+                f'New meta post: <{post.url}|{post.fullname}>'
```
perryprog: Same thing here. I'm going to assume I'm just missing something.

tor/core/inbox.py
itsthejoker: https://github.com/GrafeasGroup/tor/blob/master/tor/core/user_interaction.py#L53 This is how slack formats links. The first part is the actual link, the second part is how you want it to appear in slack. It's valid.

perryprog: Figured

PR #81, opened 2017-12-09
Allows "deno" to start the completion process

itsthejoker:
Of all the various typos we see, "deno" is by far the most popular way to butcher "done". This minor update allows this to process as normal.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -77,30 +77,33 @@ def process_reply(reply, config):
             reply.mark_read()
             return
 
-        if 'i accept' in reply.body.lower():
+        r_body = reply.body.lower()  # cache that thing
+
+        if 'i accept' in r_body:
             process_coc(reply, config)
             reply.mark_read()
             return
 
-        if 'claim' in reply.body.lower():
+        if 'claim' in r_body:
             process_claim(reply, config)
             reply.mark_read()
             return
 
         if (
-            'done' in reply.body.lower() or
-            'deno' in reply.body.lower()  # we <3 u/Lornescri
+            'done' in r_body or
+            'deno' in r_body  # we <3 u/Lornescri
         ):
-            process_done(reply, config)
+            alt_text = True if 'done' not in r_body else False
+            process_done(reply, config, alt_text_trigger=alt_text)
             reply.mark_read()
             return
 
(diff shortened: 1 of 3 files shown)


perryprog:
100% unnecessary, but it'd be so funny if we had a special response for people who misspelled it. Maybe add a dict of misspellings, and another if branch, idk.

Total waste of time, but it'd be worth it 😄

TheLonelyGhost:
In all seriousness, I'd love to get to the point where we could reply to a message like this with "I think you meant 'done'. _\[standard done response\]_". The code in its current state doesn't let us do that too easily though.

tor/core/inbox.py
```diff
@@ -87,7 +87,10 @@ def process_reply(reply, config):
             reply.mark_read()
             return
 
-        if 'done' in reply.body.lower():
+        if (
+            'done' in reply.body.lower() or
+            'deno' in reply.body.lower()  # we <3 u/Lornescri
+        ):
```
TheLonelyGhost: I was going to suggest we use a regex because it would be faster, but I just benchmarked it. Looks like running it with a regex is as much as 10x slower than 2 full text searches like this.

This should be fine. 👍

itsthejoker:
@TheLonelyGhost I just pushed up some additional changes -- will need additional review. Not too much changes computationally, just some caching and rolling in your feedback from above :)

tor/core/inbox.py
```diff
@@ -77,27 +77,33 @@ def process_reply(reply, config):
...
             reply.mark_read()
             return
 
-        if 'done' in reply.body.lower():
-            process_done(reply, config)
+        if (
+            'done' in r_body or
+            'deno' in r_body  # we <3 u/Lornescri
+        ):
+            alt_text = True if 'done' not in r_body else False
```
perryprog: Should this be changed? We're checking for done in `r_body` twice. Maybe break up the if statement or something.

Not really a big deal I guess.

itsthejoker: They do different things. One is part of the generic "do we follow this path" logic, the second is "no matter how many alternate routes we have, if that route we took to get here is not `done` then set a boolean".

perryprog: Oh yeah good point.

perryprog:
And don't forget the changelog entry of corse.

PR #84, opened 2017-12-10
fix missing import

arfie:


perryprog:
Huh,  I saw this in PyCharm and I just figured it was flaking out. Good catch.

PR #88, opened 2017-12-15
Check that the person saying "done" is the same person who claimed the post

### LINKED ISSUE #87: Validate author of "claim" is the same as author for "done"
Right now we're a bit lacking in some of our validations, specifically that the user saying "done" is the person who claimed the post originally. It's on the honor system right now. That makes me uncomfortable.

Your mission, should you choose to accept it, is to check the "claim" comment that had the bot respond with "the post is yours" (or similar) for the author and compare it to the author of a comment saying "done". If they match, continue with existing functionality. If they do not, error out and comment saying something like "you haven't claimed this post".

perryprog:
Closes #87 

Not sure if this is correct, I rushed when making it.

Also not sure if we want to use Redis to keep track of this.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import random
 
@@ -95,7 +96,7 @@ def process_claim(post, config):
             config.redis.hset(
                 'current_claims',
                 top_parent.id,
-                (post.author.name, post.permalink)
+                json.dumps([post.author.name, post.permalink])
             )
             logging.info(
                 f'Claim on ID {top_parent.fullname} by {post.author} successful'
@@ -173,10 +174,10 @@ def process_done(post, config, override=False, alt_text_trigger=False):
             # noinspection PyUnresolvedReferences
             try:
 
-                redis_result = config.redis.hget(
+                redis_result = json.loads(config.redis.hget(
                         'current_claims',
                         top_parent.id
-                )
+                ))
 
                 if redis_result[0] == post.author.name:
 
@@ -187,11 +188,21 @@ def process_done(post, config, override=False, alt_text_trigger=False):
                         ))
                     else:
                         post.reply(_(done_completed_transcript))
+
+                    # "done" was legit, so do the normal marking of stuff
                     update_user_flair(post, config)
                     logging.info(
                         f'Post {top_parent.fullname} '
                         f'completed by {post.author}!'
                     )
+
+                    config.redis.hdel(
+                        'current_claims',
+                        top_parent.id
+                    )
+                    flair_post(top_parent, flair.completed)
+
+                    config.redis.incr('total_completed', amount=1)
                 else:
                     post.reply(_(
                         done_did_not_claim_post.format(redis_result[1])
@@ -204,14 +215,6 @@ def process_done(post, config, override=False, alt_text_trigger=False):
                     f'Attempted to mark post {top_parent.fullname} '
                     f'as done... hit ClientException.'
                 )
-
-            config.redis.hdel(
-                'current_claims',
-                top_parent.id
-            )
-            flair_post(top_parent, flair.completed)
-
-            config.redis.incr('total_completed', amount=1)
 
     except praw.exceptions.APIException as e:
         if e.error_type == 'DELETED_COMMENT':
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -94,8 +94,8 @@ def process_claim(post, config):
             flair_post(top_parent, flair.in_progress)
             config.redis.hset(
                 'current_claims',
-                post.author.name,
-                (post.id, post.permalink)
+                top_parent.id,
+                (post.author.name, post.permalink)
             )
             logging.info(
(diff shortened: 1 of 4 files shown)


perryprog:
Seems like I messed up the commit history slightly ¯\_(ツ)_/¯

tor/core/user_interaction.py
```diff
@@ -166,17 +172,30 @@ def process_done(post, config, override=False, alt_text_trigger=False):
                 logging.info('Moderator override starting!')
             # noinspection PyUnresolvedReferences
             try:
-                if alt_text_trigger:
+
+                redis_result = config.redis.hget(
```
itsthejoker: hget / hset does not allow the same person to claim more than one at a time, but there is no language or error that reflects that.

perryprog: I didn't think of that, I'll see if I can fix that this weekend (or maybe sooner). I guess I should flip the key and field?

TheLonelyGhost: If we flip the hash table key and value so it's keyed by the post id, I feel like that should resolve it. Since the author name is immutable, I don't foresee any issues with user configuration making them lose their claims either.

perryprog: :+1:

perryprog:
There we go! Sorry it took so long.

tor/core/user_interaction.py
```diff
@@ -185,6 +204,11 @@ def process_done(post, config, override=False, alt_text_trigger=False):
                     f'Attempted to mark post {top_parent.fullname} '
                     f'as done... hit ClientException.'
                 )
+
+            config.redis.hdel(
```
itsthejoker: In the current configuration, this will fire if someone else attempts to complete the post.

itsthejoker:
We need to look at pickling the data or something, because in its current configuration this will not work.

```python
>>> r.hset('current_claims', 'post_id', ('post_author', 'post_permalink'))
1L
>>> r.hget('current_claims', 'post_id')
"('post_author', 'post_permalink')"
```
Data comes out as a string and I don't want to `eval` or similar in here if we don't have to -- this is a great opportunity for something like `pickle` because it's all internal.

TheLonelyGhost:
I have another idea I'd like to discuss: What if we set a time limit of, say, 10 hours for someone's claim to give the user exclusivity?

(Don't write this yet, let's discuss it first)

Instead of setting a key in a hashmap in redis, we just create a collection named `claim_{reddit_id}` containing the post permalink and the author's username. That collection could have an expiration of 10 hours from the time it's set. To check if something is claimed, try getting the value of the collection at `claim_{reddit_id}` and, if none is found, it can be assumed it is either unclaimed (which reddit flair can verify) or exclusivity time limit has been reached. If one is found, compare the author saying it was done to the person who had exclusivity.

Here are some obstacles we'd still have to work around:

- Redis expiring the `claim_*` collection wouldn't help with marking the post as unclaimed in the reddit flair. We'd have to monitor it with a bot and update the ToR post if it expires.
- Moderator support would be more difficult if someone needs to "unclaim" a post. An interface would have to be built or augmented to handle these changes.

---

Really it seems we're pretty limited unless we decide to store the state of all of the posts in redis (or a relational db), which would be its own huge PR.

itsthejoker:
@TheLonelyGhost I like that idea, but I think that would be better off for a future PR. Let's stick with the functionality we need first.

itsthejoker:
@perryprog so... do we want to wrap this up?

itsthejoker:
There's the conflict that needs to be resolved and also the fact that the current implementation of hset doesn't work; see https://github.com/GrafeasGroup/tor/pull/88#issuecomment-366765912. We'll need to find a new way to pass the data into redis so that it comes out in a way we can use.

DavidArchibald:
Because unclaim has now been added, it would also make sense to check if the person who says "unclaim" actually has claimed the post. This seems like a much more pressing matter to fix than for someone finishing something when they didn't claim it, because it's much more destrucive. This could potentially be split into another issue, but since everything being written in this PR works towards this, it seems sensible to add on.

On how this PR is going though, Pickle doesn't seem like something that should be used for turning a tuple into a string and back, you say you don't want to use an eval, but Pickle can basically be used as an eval. It's [documentation](https://docs.python.org/3/library/pickle.html) even says:

> **Warning:** The `pickle` module is not secure against erroneous or maliciously constructed data. Never unpickle data received from an untrusted or unauthenticated source.

Which is the commonly cited as the main problem with eval. I'd suggest something like csv instead if that's your concern.

itsthejoker:
@DavidArchibald This PR is on hold because it makes a lot more sense to implement that functionality after the database is done, and that's the topic of this holiday break.

Also, in this case, the source of the pickled string is always trusted, so pickle is fine.

TheLonelyGhost:
We're deferring this functionality until we have a SQL database backing the bots (as opposed to the current hacky use of Redis). Once that's done, it will need to be almost completely reimplemented, let alone the merge conflicts that will need to be addressed if we tried to carry this PR forward.

Not that it's a bad PR or that it's a bad implementation, the bot just needs a few renovations before it'd be ready for this.

PR #89, opened 2017-12-17
History checker blows up on pinned submissions

TheLonelyGhost:
If a user actually uses their feed for more than just comment history (it's possible nowadays to have your own subreddit-like experience on your user profile), we skip it. This became necessary when we wouldn't respond to `done` with the new history checker functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,6 +81,7 @@ setup(
     install_requires=[
         'tor_core',
         'sh',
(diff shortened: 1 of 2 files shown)


TheLonelyGhost:
Also bumped up the number of comments to inspect to 10, due to some of them possibly being stickied submissions.

PR #92, opened 2018-01-15
Closes #90. Update some interaction language and slack messages.

### LINKED ISSUE #90: a(n) image post :(
Changing the text in u/transcribersofreddit's claim info post from-

"This is a(n) image post, so please use the following formatting:"

to 

"Post type: image. Please use the following formatting:

itsthejoker:


tor/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '3.4.0'
+__version__ = '3.5.0'
```
perryprog: Shouldn't this be 3.4.1? Semvar rules, and everything?

itsthejoker: Look at what it's being merged into ;) While yes, normally I would actually leave this change undone, we're switching to feature branches -- that's why your PR was moved against the 3.5.0 release branch :)

itsthejoker: But since you brought it up, I'll go ahead and roll that back.

perryprog: Oh—I didn't see that.

yeah, probably best to have that in the final PR.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 We follow [Semantic Versioning](http://semver.org/) as a way of measuring stability of an update. This
 means we will never make a backwards-incompatible change within a major version of the project.
 
+## Unreleased (2018-1-15)
```
TheLonelyGhost: If we're gonna have it be in an "unreleased" section, we should avoid the datestamping here. Think about it this way:

> "When was this 'Unreleased'? Oh, on that day? What about after that date: was it released then?"

The datestamping on the ones listed below with version numbers are just a convenience thing. They are a snapshot in time and we're noting when the snapshot was taken along with the semantic versioning style of version number.

tor/helpers/flair.py
```diff
@@ -142,6 +142,8 @@ def set_meta_flair_on_other_posts(config):
             )
             flair_post(post, flair.meta)
             send_to_slack(
-                f'New meta post: <{post.url}|{post.fullname}>',
+                f':thinking: '
+                f'New meta post: <{post.url}|{post.submission.shortlink}> '
```
TheLonelyGhost: I get that `post.fullname` just outputs the internal PRAW id, but what about the post title instead? We're already making this clickable as a link, wouldn't it be more helpful to annotate the link with the title of the post?

arfie: I originally meant to put the title when I wrote this, but got those mixed up somehow. +1

itsthejoker: Ah, crap. I totally forgot to change that -- I'll change it to title instead.

PR #94, opened 2018-01-20
Update user commands structure

arfie:
Got a "done" comment today where the transcriber had added the URL of their transcription - which happened to include the word "fuck" so the "done" was not processed.

https://www.reddit.com/r/TranscribersOfReddit/comments/7rs9yq/trollxchromosomes_image_fuckboys_be_like/dszh158/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 We follow [Semantic Versioning](http://semver.org/) as a way of measuring stability of an update. This
 means we will never make a backwards-incompatible change within a major version of the project.
 
+## Unreleased
+
+- Fix mod interventions to not prevent normal user commands from being executed (by @arfie)
+- Change the user commands structure (by @arfie)
+
 ## v3.4.0 (2017-12-11)
 
 - Send inbox messages to Slack if they don't match any of the commands (by @arfie)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 We follow [Semantic Versioning](http://semver.org/) as a way of measuring stability of an update. This
 means we will never make a backwards-incompatible change within a major version of the project.
 
-## Unreleased (2018-1-25)
+## Unreleased
 
 - Verbiage updates and other message improvements (by @itsthejoker)
 - Fix mod interventions to not prevent normal user commands from being executed (by @arfie)
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -12,6 +12,7 @@ from tor.core.user_interaction import process_claim
 from tor.core.user_interaction import process_coc
 from tor.core.user_interaction import process_done
 from tor.core.user_interaction import process_thanks
+from tor.core.user_interaction import user_commands
 
 MOD_SUPPORT_PHRASES = [
     re.compile('fuck', re.IGNORECASE),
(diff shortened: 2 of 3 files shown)


TheLonelyGhost:
@perryprog it’s because we don’t take kindly to that meme. Not sure if it’s something we’ll just ignore in the future, but for now we’re letting kids know.

tor/core/inbox.py
```diff
@@ -75,7 +75,6 @@ def process_reply(reply, config):
         if any([regex.search(reply.body) for regex in MOD_SUPPORT_PHRASES]):
             process_mod_intervention(reply, config)
             reply.mark_read()
```
TheLonelyGhost: I would also remove this line so that there wouldn’t be issues if we try to mark it as read twice.

I imagine we would want the full messages caught in the unhandled message handler that make it through here.

TheLonelyGhost:
Good suggestion! It didn’t even occur to me that this would be an issue.

perryprog:
Ya, it was a problem a couple weeks ago with that. Someone did a huge transcription and was like '****ing done!' or something similar at least.

arfie:
I've changed it so replies will only be marked read once, and mod interventions will not also be sent as unrecognized messages (which would've happened after my previous commit).

On top of that, I'd like some review on my latest commit: I've changed the user command system to register commands with a decorator, rather than having to add it to the ugly if-elif chain, saving on some duplicate code. I originally had this idea for the admin command overhaul, but never got to that and Perry's system works as well. The main advantage of this is that adding user commands now only requires change in one place in one file, and there's little duplicate code any more.

tor/core/user_interaction.py
```diff
@@ -35,6 +35,24 @@ def coc_accepted(post, config):
     return config.redis.sismember('accepted_CoC', post.author.name) == 1
 
 
+user_commands = {}
```
itsthejoker: I have to get to my desktop to test this, but as I'm reading this is seems like we're importing this dict and never filling it. Am I reading this wrong?

tor/core/inbox.py
```diff
@@ -12,6 +12,7 @@
 from tor.core.user_interaction import process_coc
 from tor.core.user_interaction import process_done
 from tor.core.user_interaction import process_thanks
```
perryprog: All of these imports are now unused

tor/core/user_interaction.py
itsthejoker: Thanks, I've never seen it done that way and I missed that. I've tested it manually and this is probably one of the cleanest changes we've ever had to the codebase. :tada: :tada:

tor/core/inbox.py
```diff
@@ -77,38 +78,23 @@ def process_reply(reply, config):
...
-            handled = True
-
-        if not handled:
-            # If we made it this far, it's something
-            # we can't process automatically
-            forward_to_slack(reply, config)
+        for triggers, cmd in user_commands.items():
+            if any(t in r_body for t in triggers):
+                cmd(reply, config)
+                break
```
TheLonelyGhost: Wouldn't this require `user_commands` to be an ordered dict, since the block above quits after finding the first matching trigger?

Let's say I have `"done. good bot. unclaim"`. Which of the 3 branches would be evaluated and the rest thrown away? I'm not sure I'd be able to predict it.

tor/core/user_interaction.py
TheLonelyGhost: As @itsthejoker mentioned earlier in this thread, using decorators adds a level of indirection to the code that would make tracing it a bit more difficult. That said this is a lot clearer in what we're trying to do overall by hiding the implementation details. I like it. 👍

itsthejoker:
Oh, almost forgot! Can you please change the merge branch from `master` to `3.5.0` and update CHANGELOG? :)

perryprog:
Might be worthwhile to move admin_commands over to a decorator thing too, with detailed params in the decorator on how they should be called and such.

For example:
```py
@admin_command(whitelist=['personal_opinions'], pass_full_message=True)
```

Or something similar.

itsthejoker:
@perryprog while that is a good idea, it wouldn't be for this PR. Do you want to make an issue for it?

tor/core/inbox.py
arfie: Good catch!

tor/core/inbox.py
arfie: It works with a dict, but the order is indeed not really predictable/changeable. Since we also don't need random access, an ordereddict will indeed work better :+1:

CHANGELOG.md
```diff
@@ -3,8 +3,9 @@
 We follow [Semantic Versioning](http://semver.org/) as a way of measuring stability of an update. This
 means we will never make a backwards-incompatible change within a major version of the project.
 
-## Unreleased
+## Unreleased (2018-1-25)
```
TheLonelyGhost: Why are we adding a date when it was... unreleased?

PR #100, opened 2018-02-03
Change history validation to only look for comments

arfie:
So apparently the history validation doesn't work properly when someone has submissions stickied to their profile. I was unable to replicate this, but it should be fixed anyway by this.


perryprog:
Happy 100th Pull Request/Issue on this repo!! 🎉 🎊

Issue #102, opened 2018-02-15
Allow changing verbiage of claim success on CoC acceptance


itsthejoker:
Currently when a new user comes on board, they go through the code of conduct and that simply calls the rest of the claim code. The current message reads,

```
The post is yours! Best of luck and thanks for helping!

Please respond with "done" when complete so we can check this one off the list!
```

Let's add a flag to that so we know if we're coming from the code of conduct so we can tweak that message to make it more friendly and tell people to talk to the mods if they get stuck. Maybe something like,

```
The post is yours! Best of luck and thanks for helping!

Since this is your first one, please make sure to message the mods if you have any questions about the process (and / or join our [Discord](link) to talk to other volunteers). We're here to help!

Please respond with "done" when complete so we can check this one off the list!
```

itsthejoker:
And maybe mention the formats again? From @pejmanpoh

PejmanPoh:
I might take a crack at this to get acquainted with the code base. Seems like tor_app is a long way from seeing any code. Also a good chance to start using python anyway.

DavidArchibald:
As a (very) new member of /r/ToR, this idea seems like a good idea to help new transcribers along. The mod support offered should help, plus the discord link is even better as I missed it in my reading of the sidebar, likely because I skimmed over some of the first parts though... it seemed pretty standard. (I think other subreddits I've seen have have put their discord link lower on the sidebar and more bold, but it's probably my fault.)

When I came into the subreddit, I was a bit unsure what to do, even with the template and wiki pages. Ex. Remove the square brackets or not(they have meant insertion in almost every other non coding thing I've done).

My point in sharing that is that I didn't think I could do anything, but hope that I wasn't about to be temp banned for 7 days for messing up the formatting(which I genuinely thought could happen, but was soon dispelled). 

So while in my case a nice mod came and verified I'd done it okay, telling the new redditors they can contact mods could help.

itsthejoker:
Bumping this as it's come up as a request again from the welcome team.

PR #104, opened 2018-02-16
First claim message

PejmanPoh:
Addresses issue #102

I couldn't think of an elegant way of implementing a flag from the `process_coc` without adding a new parameter to `process_claim`. Instead, I've mocked up a flair count check to verify if it is the user's first claim.

- Added a new function to get flair count (Could find an implementation of this, nor any post.get_author_flair in PRAW? If this function is fine, I can integrate it into the other `flair.py` functions)
- Added a new reply message for a user's first claim

Please be extra cautious for PEP 8 errors. First time I haven't had to make up styling on the spot.


tor/strings/responses.py
```diff
@@ -9,8 +9,16 @@
 )
 
 claim_success = (
-    'The post is yours! Best of luck and thanks for helping!\n\nPlease respond '
-    'with "done" when complete so we can check this one off the list!'
+    'The post is yours! Best of luck and thanks for helping!\n\n {} Please '
```
itsthejoker: This line will have 2 spaces before "Please" if there's nothing provided. Here you can just take the spaces out on either side of the {}.

tor/core/user_interaction.py
```diff
@@ -12,6 +12,7 @@
 from tor.helpers.flair import flair
 from tor.helpers.flair import flair_post
 from tor.helpers.flair import update_user_flair
+from tor.helpers.flair import get_user_flair_count
```
itsthejoker: do we still need this import?

tor/strings/responses.py
PejmanPoh: :+1:

tor/core/user_interaction.py
```diff
@@ -88,7 +88,10 @@ def process_claim(post, config):
 
         if flair.unclaimed in top_parent.link_flair_text:
             # need to get that "Summoned - Unclaimed" in there too
-            post.reply(_(claim_success))
+            if first_claim:
+                post.reply(_(claim_success.format(first_claim_success)))
+            else:
+                post.reply(_(claim_success))
```
itsthejoker: We'll want to format this with an empty string, otherwise the `{}` will show up in the final product. Example:

```python
>>> x = "Test!{}Test!"
>>> print(x)
Test!{}Test!
>>> print(x.format(''))
Test!Test!
>>>
```

PejmanPoh: Ah good point :+1:

tor/core/user_interaction.py
```diff
@@ -88,7 +88,10 @@ def process_claim(post, config):
 
         if flair.unclaimed in top_parent.link_flair_text:
             # need to get that "Summoned - Unclaimed" in there too
-            post.reply(_(claim_success))
+            if first_claim:
+                post.reply(_(claim_success.format(first_claim_success)))
+            else:
+                post.reply(_(claim_success.format(''))
```
TheLonelyGhost: I’m not a huge fan of shoving one template into another like this since it encourages placing too much logic in a string formatted when it would probably fit better in a method call that follows a string builder pattern.

In this case, though, I think it’s simple enough that it might be clearer to have some duplication. Perhaps we could have 2 separate strings be used for the messages?

```python
if first_claim:
    post.reply(_(first_claim_success))
else:
    post.reply(_(claim_success))
```

tor/core/user_interaction.py
```diff
@@ -53,10 +53,10 @@ def process_coc(post, config):
             f'<http://www.reddit.com/u/{post.author.name}|u/{post.author.name}>'
             f' has just accepted the CoC!', config
         )
-    process_claim(post, config)
+    process_claim(post, config, first_claim=True)
```
TheLonelyGhost: I like that we’re setting this as a keyword argument instead of a positional one. Definitely reveals the intention a lot more clearly. :+1:

tor/core/user_interaction.py
```diff
@@ -88,7 +88,10 @@ def process_claim(post, config):
 
         if flair.unclaimed in top_parent.link_flair_text:
             # need to get that "Summoned - Unclaimed" in there too
-            post.reply(_(claim_success))
+            if first_claim:
+                post.reply(_(claim_success.format(first_claim_success)))
```
TheLonelyGhost: (might be a moot point with the other comments I left, but...)

Looks like we're missing an import for `first_claim_success`

tor/core/user_interaction.py
PejmanPoh: It was @itsthejoker decision to switch to this 'shoving' method. What do you think Joe?

tor/core/user_interaction.py
PejmanPoh: :+1:

tor/core/user_interaction.py
TheLonelyGhost: To be clear, we should be keeping the first-claim messaging almost the same as the subsequent claim messages, but I'm not sure we should be enforcing that with code. That's more of an opinion thing though. Also interested in hearing Joe's thoughts on the matter.

PejmanPoh: @itsthejoker ping

itsthejoker: Ugh, sorry. I personally prefer logic-ing the templates because of DRY. I'd rather update one template and not have to worry about it than updating two. That said, this is short enough that it's probably not an issue.

TheLonelyGhost: You know, I'm not gonna [bikeshed](https://en.wiktionary.org/wiki/bikeshedding) this. It's low enough importance that we can just keep it as-is. 👍

tor/core/user_interaction.py
TheLonelyGhost: Still needs the missing import. Then I'll approve the PR.

itsthejoker:
Along with what @TheLonelyGhost mentioned about the import, you also have to resolve the conflicts :)

PR #107, opened 2018-03-11
update verbiage of rules comment

itsthejoker:
Rearrange the rules comment a little bit. Had a newbie get confused by the current layout, so this fixes that issue.

https://www.reddit.com/r/TranscribersOfReddit/comments/83kmp4/me_irl_image_me_irl/dviph2j/?context=3&utm_content=context&utm_medium=message&utm_source=reddit&utm_name=frontpage

### DIFF
--- a/tor/strings/posts.py
+++ b/tor/strings/posts.py
@@ -14,8 +14,7 @@ rules_comment = (
     'wiki/checklist) that might help you out -- not just for newbies!'
     '\n\nThis is a(n) {post_type} post, so please use the following formatting:'
     '\n\n---\n\n{formatting}\n\n---'
-    '\n\nFooter\n---\n\nWhen you\'re done, please put the following footer at '
-    'the **bottom** of your post:\n\n---\n\n{header}\n\n---'
+    '\n\nReporting\n---'  # this is a header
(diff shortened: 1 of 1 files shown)


tor/strings/posts.py
```diff
@@ -14,15 +14,18 @@
     'wiki/checklist) that might help you out -- not just for newbies!'
     '\n\nThis is a(n) {post_type} post, so please use the following formatting:'
     '\n\n---\n\n{formatting}\n\n---'
-    '\n\nFooter\n---\n\nWhen you\'re done, please put the following footer at '
-    'the **bottom** of your post:\n\n---\n\n{header}\n\n---'
+    '\n\nReporting\n---'  # this is a header
```
arfie: wouldn't it be better to use `## Reporting` style headers in this case? The underline style seems to only be for visual purposes, which doesn't really work in this case.

Or we could wrap the whole thing in a `"""multi line comment"""` which would also solve having to insert all the apostrophes and `\n`s. Idk it's only a matter of code style

itsthejoker: Yes, that is a change we should eventually make. I think it would be good to go through and change all our messages to multiline comments eventually, but that should be its own thing.

PR #117, opened 2018-04-14
update code of conduct channel and make message more fun

itsthejoker:
Move new volunteer notifications to their own channel. And make the message more fun.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -46,12 +46,33 @@ def process_coc(post, config):
     """
     result = config.redis.sadd('accepted_CoC', post.author.name)
 
+    modchat_emote = [
+        ':tada:',
+        ':confetti_ball:',
+        ':party-lexi:',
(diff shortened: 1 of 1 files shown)


tor/core/user_interaction.py
```diff
@@ -46,12 +46,33 @@ def process_coc(post, config):
...
+        ':fb-like:'
+    ]
+
     # Have they already been added? If 0, then just act like they said `claim`
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_modchat(
             f'[u/{post.author.name}](http://www.reddit.com/u/{post.author.name})'
-            f' has just accepted the CoC!', config
+            f' has just accepted the CoC! {modchat_emote}',
```
TheLonelyGhost: The variable `modchat_emote` is an list datatype instead of a string. Do you mean to use `random.choice(modchat_emote)` instead?

PR #124, opened 2018-06-06
Flair fix

itsthejoker:


tor/core/users.py
```diff
@@ -0,0 +1,116 @@
...
+        self.user_data.update({'username': self.username})
+        return self.user_data
+
+
+if __name__ == '__main__':
+    pam = User('pam')
+    print(pam)
+    pam.update('transcriptions', pam.get('transcriptions', 1) + 1)
+    print(pam.get('transcriptions'))
+    pam.save()
```
TheLonelyGhost: @itsthejoker Is this section still needed? Looks like debugging code to me.

itsthejoker: this only runs if the file is run by itself. If anything it's documentation.

TheLonelyGhost: Eh we're already refactoring it with tor_worker anyway. 👍

PR #125, opened 2018-06-22
Change Check Submissions to use threads

itsthejoker:
This PR completely rebuilds the existing system that checks for new posts. Currently, running through all of our partner subreddits and running everything through the filters takes around 58 seconds.

This PR takes that down to ~1.6-2.3 seconds.

I've tested the threading section extensively, but `process_post` needs more manual testing. I am exhausted, so I'm putting this PR up here so I can hit the hay and work on this later.

The goal is to get this tested, reviewed, and merged before the Clear the Queue event on the 23rd.

### DIFF
--- a/tor/helpers/threaded_worker.py
+++ b/tor/helpers/threaded_worker.py
@@ -1,0 +1,116 @@
+# After a lengthy discussion with @thelonelyghost, it became apparent that we're
+# fully expecting to completely throw away this codebase when we finish the
+# rewrite for celery. Therefore, ugly-ass hacks are allowed in the name of
+# making the current bots faster if at all possible. Is it maintainable?
+# Absolutely not, but that's a conscious decision on our part. Feast your eyes
+# upon the ugliness... and know that we are sorry.
+
+import logging
+import random
+import string
+from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import as_completed
+
+import requests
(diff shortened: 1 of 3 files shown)


tor/helpers/threaded_worker.py
```diff
@@ -0,0 +1,114 @@
...
+        for sub in subreddits:
+            jobs.append(executor.submit(get_subreddit_posts, sub))
+        for f in as_completed(jobs):
+            try:
+                data = f.result()
+                total_posts += data
+            except Exception as exc:
+                logging.warning('an exception was generated: {}'.format(exc))
+
+    for item in total_posts:
```
arfie: Is there a reason why we're waiting for all threads to complete before processing the posts? Wouldn't it be nicer and more efficient to do this as soon as the jobs finish?

Also, I think it'd probably be nicer to move the domain check  into `process_post` as other filtering also occurs there. Maybe even move the entire filter process into its own method?

itsthejoker: I like the idea to move the filtering to its own method -- that'd probably make it clearer.

As for doing this as soon as the jobs finish, that's definitely possible but it's more complicated. Doing it this way is the recommended way in the Python3 documentation for guaranteeing that all the threads get cleaned up properly. Less work on our side.

PR #126, opened 2018-06-23
Add in some delay for checking for new posts

itsthejoker:
We sped up checking for new posts so much that we're kind of hammering Reddit's API right now. This change makes it so that we only check for new posts once every 45 seconds and has a matching PR in tor_core.

### DIFF
--- a/tor/helpers/threaded_worker.py
+++ b/tor/helpers/threaded_worker.py
@@ -8,6 +8,8 @@
 import logging
 import random
 import string
+from datetime import timedelta
+from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import as_completed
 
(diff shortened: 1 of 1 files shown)


tor/helpers/threaded_worker.py
```diff
@@ -95,6 +97,13 @@ def parse_json_posts(posts: Dict) -> List:
     return parse_json_posts(result)
 
 
+def is_time_to_scan(config) -> bool:
+    if datetime.now() > config.last_post_scan_time + timedelta(seconds=45):
+        config.last_post_scan_time = datetime.now()
```
TheLonelyGhost: Having a method named like this would make me think it does not mutate any variable. I see why it was made this way, but I’d suggest separating them out.

Additionally, given how the threads work with python, I’m concerned using config would result in each thread modifying shared memory. Have you looked into registering a global, thread-local variable explicitly?

itsthejoker: Good point about the method name, I'll fix that.

For threads, this section is not run by the threads. The ThreadPoolExecutor on line 128 is what kicks them off -- when we process this section, we're still inside the primary thread. The individual threads only handle the code referenced from https://github.com/GrafeasGroup/tor/blob/994740b1c68310bdb5e05e1ed6b857697dc64616/tor/helpers/threaded_worker.py#L128-L137, which doesn't include the `config` object.

PR #128, opened 2018-06-27
Fix no method on nonetype

TheLonelyGhost:
Fixes issue on bugsnag: "Argument of type NoneType is not iterable" with `tor/core/user_interaction.py:114`

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -103,15 +103,18 @@ def process_claim(post, config):
             ))
             return
 
+        # Because praw doesn't cache this calculated property. We reload it each time.
+        flair_text = top_parent.link_flair_text
+
         # this can be either '' or None depending on how the API is feeling
         # today
-        if top_parent.link_flair_text in ['', None]:
+        if flair_text in ['', None]:
             # There exists the very small possibility that the post was
             # malformed and doesn't actually have flair on it. In that case,
             # let's set something so the next part doesn't crash.
             flair_post(top_parent, flair.unclaimed)
 
-        if flair.unclaimed in top_parent.link_flair_text:
+        if flair.unclaimed in flair_text:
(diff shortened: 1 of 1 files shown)


tor/core/user_interaction.py
```diff
@@ -103,15 +103,18 @@ def process_claim(post, config):
...
         # today
-        if top_parent.link_flair_text in ['', None]:
+        if flair_text in ['', None]:
             # There exists the very small possibility that the post was
             # malformed and doesn't actually have flair on it. In that case,
             # let's set something so the next part doesn't crash.
             flair_post(top_parent, flair.unclaimed)
 
-        if flair.unclaimed in top_parent.link_flair_text:
+        if flair.unclaimed in flair_text:
```
itsthejoker: This will cause an issue if Reddit ever returns an empty response for flair text. Let's use None for example:

Original path:

- if flair_text is none:
  - flair post. This changes the flair on Reddit's side.
- Now we pull it again -- does the flair have the word "unclaimed" in it?
  - if so, process.

PR process:
- cache flair_text -- it won't be changed for the duration of this segment.
- if flair_text is none:
  - flair post. This changes the flair on Reddit's side.
- Does the cached flair have the word "unclaimed" in it?
  - Nope, it's still None, so it won't get processed.

TheLonelyGhost: I didn't think about that. It sounds like we're trying to work around issues with Reddit being inconsistent in general, which is unacceptable if it's our source of truth.

Perhaps having our own data store tracking the status of transcribable posts would be a better idea, then replicating the changes to the data store over to Reddit similar to how we're tracking user information now.

TheLonelyGhost:
After further thought, this is trying to "fix" desired functionality, which flares up because using Reddit as a data store is very inconsistent in its availability.

PR #130, opened 2018-06-29
Adds GitLab CI integration for linting and pytest

TheLonelyGhost:
I'm totally open to suggestions, but it looks like this would make it quick and easy to just use GitLab's CI component, even to report back in on GitHub.

Some things to note are that all of the linting or static analysis tools (`black`, `pylint`, and `flake8`) are running without configuration files. That can be changed pretty easily.

Additionally, this runs under Python 3.7.0b5 only because the released version of Python 3.7.0 is not yet available as a Docker image, per docker-library/python#297.

Given that this is run on a Docker container, one can troubleshoot an issue locally using the publicly available image. Much nicer than Travis or Circle CI!

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,38 @@
+---
+image: 'python:3.7-rc-alpine'
+
+variables:
+  GIT_DEPTH: "1"
+
+stages:
+  - linting
+  - tests
+
+black:
+  stage: linting
+  before_script:
+    - 'pip3 install black'
+  script:
+    - 'black --check setup.py tor'
+
+flake8:
+  stage: linting
+  before_script:
+    - 'pip3 install flake8'
+  script:
+    - 'flake8 tor'
+
+pylint:
+  stage: linting
+  before_script:
+    - 'pip3 install pylint'
+  script:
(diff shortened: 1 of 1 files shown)


TheLonelyGhost:
Just tested again with pushing an update. Slight issue found: Syncing is far from immediate and the waiting period doesn't even mention the gitlab-ci status checks, let alone that they might be "pending".

Some pending:

completed WIP and pending Travis-CI checks = 1/3 successful checks

Still no GitLab CI checks:

completed WIP and Travis-CI checks = 3/3 passing checks

Unless there's a way to initiate syncing the repositories, (e.g., webhook on every push to the repo) it looks like GitLab CI is a nice-to-have but not to be relied upon. :disappointed:

TheLonelyGhost:
Eventually:

screenshot_2018-07-04_14-20-53

Final result:

screenshot_2018-07-04_14-35-30

It looks like there's an option for [mirroring repositories by API call](https://docs.gitlab.com/ee/api/projects.html#start-the-pull-mirroring-process-for-a-project), so I'm going to try triggering that from Travis' build process. Perhaps that will get us to the pending state faster...

Issue #121, opened 2018-05-23
Add link to accept CoC post


PejmanPoh:
Currently, the RC notification simply states that

    /u/XYZ has just accepted the CoC!

It would be useful to have a direct link to said `accept` message.

PR #135, opened 2018-08-19
Add direct links to send_to_modchat messages

PejmanPoh:
Issue #121

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -72,7 +72,7 @@ def process_coc(post, config):
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_modchat(
-            f'<{reddit_url.format('/user/' + post.author.name)}|u/{post.author.name}>'
+            f'<{reddit_url.format("/user/" + post.author.name)}|u/{post.author.name}>'
             f' has just'
             f'<{reddit_url.format(post.permalink)}|accepted the CoC!>'
             f' {modchat_emote}',
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -23,6 +23,7 @@ from tor_core.helpers import clean_id
 from tor_core.helpers import get_parent_post_id
 from tor_core.helpers import get_wiki_page
 from tor_core.helpers import send_to_modchat
+from tor_core.strings import reddit_url
 
 
 def coc_accepted(post, config):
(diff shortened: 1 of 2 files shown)


tor/core/user_interaction.py
```diff
@@ -71,8 +72,10 @@ def process_coc(post, config):
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_modchat(
-            f'<http://www.reddit.com/u/{post.author.name}|u/{post.author.name}>'
-            f' has just accepted the CoC! {modchat_emote}',
+            f'<{reddit_url.format(\'/user/\' + post.author.name)}|u/{post.author.name}>'
```
itsthejoker: what's going on here?

itsthejoker: Just needs to be `reddit_url.format('/user/' + post.author.name)`. No escaping necessary :)

PejmanPoh: Haha never seen highlighting red like that. Removed the escaping.

arfie: Not quite, because then the string literal would be ended (since it already uses single quotes). The best option is to change to double quotes in the fstring expression, ie. `reddit_url.format("/user/" + post.author.name)`.

tor/core/user_interaction.py
```diff
@@ -71,8 +72,10 @@ def process_coc(post, config):
     # instead. If they're actually new, then send a message to slack.
     if result == 1:
         send_to_modchat(
-            f'<http://www.reddit.com/u/{post.author.name}|u/{post.author.name}>'
-            f' has just accepted the CoC! {modchat_emote}',
+            f'<{reddit_url.format('/user/' + post.author.name)}|u/{post.author.name}>'
```
arfie: While it looks correct according to GitHub's syntax highlighting, this also doesn't work:

```
>>> def foo(bar): return bar
... 
>>> f'{foo('bar')}'
  File "<stdin>", line 1
    f'{foo('bar')}'
              ^                                                                                                                                                                                              
SyntaxError: invalid syntax
```

I suggest using double quotes, ie. `reddit_url.format("/user/" + post.author.name)`.

itsthejoker: oops part of that is on me. I didn't correct for that. Sorry!

PejmanPoh:
@itsthejoker @arfie Updated it to double quotes

PR #137, opened 2018-09-11
Allow "dibs" to claim a post

DavidArchibald:
Just for fun add dibs as a way to claim a post.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -96,7 +96,7 @@ def process_reply(reply, config):
 
         if (
             'done' in r_body or
-            'deno' in r_body # we <3 u/Lornescri
+            'deno' in r_body  # we <3 u/Lornescri
         ):
             alt_text = True if 'done' not in r_body else False
             process_done(reply, config, alt_text_trigger=alt_text)


tor/core/inbox.py
```diff
@@ -93,7 +93,8 @@ def process_reply(reply, config):
 
         if (
             'done' in r_body or
-            'deno' in r_body  # we <3 u/Lornescri
+            'deno' in r_body or # we <3 u/Lornescri
+            'dibs' in r_body
```
itsthejoker: Putting it here would only allow it to finish a post, not claim one.

DavidArchibald: Oops it started with "d" and fit the pattern so I just forgot basic logic 😅

tor/core/inbox.py
```diff
@@ -86,14 +86,17 @@ def process_reply(reply, config):
...
+        ):
             process_claim(reply, config)
             reply.mark_read()
             return
 
         if (
             'done' in r_body or
-            'deno' in r_body  # we <3 u/Lornescri
+            'deno' in r_body # we <3 u/Lornescri
```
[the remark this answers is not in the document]
DavidArchibald: Maybe I shouldn't edit on my phone... I'm so used to black doing it for me, sorry for making this much harder than it has to be.

PejmanPoh:
👍

PR #138, opened 2018-11-12
Allow dibs to claim a post.

DavidArchibald:
I'm really sorry if I managed to mess this up like last time.

For example: According to [`CONTRIBUTING.md`](https://github.com/GrafeasGroup/tor/blob/b7725413b6c2269397f03ce3e561a57657adfebd/CONTRIBUTING.md) it should be `credit: @username` not just `@username` or `by @username`, however there is no instance of that in [`CHANGELOG.md`](https://github.com/GrafeasGroup/tor/blob/b7725413b6c2269397f03ce3e561a57657adfebd/CHANGELOG.md), so I went with @username.


tor/core/inbox.py
```diff
@@ -86,7 +86,7 @@ def process_reply(reply, config):
             reply.mark_read()
             return
 
-        if 'claim' in r_body:
+        if 'claim' in r_body or 'dibs' in r_body:
```
itsthejoker: For stylistic purposes, I'd like to keep the parentheses like with the `done` command. Makes it easier to see that there are two separate commands which trigger the same functions.

PR #139, opened 2018-11-20
Politely yell at people who comment transcriptions on the ToR post

itsthejoker:
Adds an additional check to see if the incoming message is a transcription so that it stops making its way to slack.

Also contains some slight reformatting of strings/responses.py for readability purposes.


tor/core/inbox.py
```diff
@@ -113,6 +114,15 @@ def process_reply(reply, config):
             reply.mark_read()
             return
 
+        if (
+            '&#32;' in r_body or
+            'reddit/wiki/index)' in r_body or  # the end of the footer
+            'image transcription' in r_body
+        ):
```
arfie: Why not reuse `footer_check` from `tor.core.validation`? That should work just the same here.
I like also checking for "image transcription" since people are bad at reading.

arfie: Also, and this is actually quite important: move this check up a few places. A transcription could include one of the other trigger words, especially "thank" could be problematic.

itsthejoker: 1) why not use `footer_check`: because... yeah no that's a good idea

2) yes, definitely will move it up

tor/core/inbox.py
```diff
@@ -82,6 +83,15 @@ def process_reply(reply, config):
 
         r_body = reply.body.lower()  # cache that thing
 
+        if (
+            '&#32;' in r_body or
```
arfie: This line can be left out with #140 :p

itsthejoker:
🎉

tor/core/inbox.py
itsthejoker: dammit, you're totally right

PR #140, opened 2018-11-20
make validation look for a properly formatted footer

itsthejoker:
Just adds a small check to stop people from rewriting the footer and forcing them to use the markdown abomination that is our current footer.


itsthejoker:
sounds good to me.

PR #144, opened 2018-12-21
Update slack messages a bit

arfie:
Mainly changes the meta post notification to actually link to the post rather than the attached URL. Also includes the body of messages without an author, just like with unhandled messages.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -160,8 +160,8 @@ def check_inbox(config):
         # In this case, there will be no author attribute.
         if item.author is None:
             send_to_modchat(
-                f'We received a message without an author. Subject: '
-                f'{item.subject}', config
+                f'We received a message without an author -- '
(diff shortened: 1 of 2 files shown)


tor/core/inbox.py
```diff
@@ -160,8 +160,8 @@ def check_inbox(config):
         # In this case, there will be no author attribute.
         if item.author is None:
             send_to_modchat(
-                f'We received a message without an author. Subject: '
-                f'{item.subject}', config
+                f'We received a message without an author -- '
+                f'*{item.subject}*:\n{item.body}', config
```
itsthejoker: do we want to do two `\n`'s so that we have an empty line here?

arfie: I based it on `forward_to_slack`, which uses only one newline. I think it looks fine.

itsthejoker: Roger that, works for me.

PR #145, opened 2018-12-24
Small grammar and code styles fixes.

DavidArchibald:
I noticed some small grammar errors and code style problems. This fixes lots of small things, but the most common ones are:

- Double quotes are turned into single quotes, because most of the codebase used them, but not all.

- Every sentences' beginning is capitalized and end with a period, along with other small grammar fixes/word choices.

- `x in ['foo', 'bar']` is turned into `y in ('foo', 'bar')`, because tuples are immutable and don't have the(tiny) overhead lists do... basically just boils down to consistency and style.

This shouldn't change *any* actual functions of the script. If it does, it's a mistake.

PR #146, opened 2018-12-24
Add Direct Message replies.

### LINKED ISSUE #131: Bot messages you back when you send it a direct message
At the moment, other moderators can't access /u/TOR's messages. Transcribers message the bot, and don't get a response. 

An auto response for direct messages, like when you tag /u/TOR in a comment - would be good.

DavidArchibald:
I'm not sure if having #145 open too is a problem since they both bump the version to `3.6.1`... I just feel like I messed something up somehow in general! Regardless of potential mistakes, this intends to close #131.

A note for both #131 and this PR's code, the distinction between "PM" and "DM" is a little confusing. From what I gathered in the issue, a comment with text like "foo bar /u/ToR" is a PM and messaging /u/ToR is a DM. I'd suggest revising the word choice for these, because I've usually heard PM and DM used synonymously, and PM does not make me think of mentioning a user in a message.


tor/core/user_interaction.py
```diff
@@ -328,3 +328,19 @@ def process_wrong_post_location(post):
             'Something went wrong with asking about a misplaced post; '
             'ignoring.'
         )
+
+def process_message(message, config):
+    if not isinstance(message, praw.models.Message):
```
itsthejoker: I'd just set `praw.models.Message` in line 132 as a statically typed param. `def process_message(message: praw.models.Message, config):`

tor/strings/responses.py
```diff
@@ -126,6 +126,22 @@
...
 
+dm_subject = (
+    'Direct Message'
+)
+
+dm_body = (
+    'Hi there! Thanks for messaging me!'
+    '\n\n'
+    'Unfortunately I\'m a bot and so I can\'t help you out myself. I\'ve '
+    'forwarded your message to a moderator so they can help you out! They '
```
itsthejoker: Let's change this to "forwarded your message to the moderators" -- don't want people to get confused and start PM'ing us directly trying to see if we got their message.

tor/strings/responses.py
```diff
@@ -126,6 +126,22 @@
...
+    'should message you shortly. In the future, you can [message the mods]'
```
itsthejoker: `you can [message the mods at this link]` or something along those lines.

tor/strings/responses.py
```diff
@@ -126,6 +126,22 @@
...
+    '(https://www.reddit.com/message/compose?to=%2Fr%2FTranscribersOfReddit&'
+    'subject=Transcription%20Help&message=) to talk to them directly.'
```
itsthejoker: `to reach them directly`. Has a bit more of a formal tone.

DavidArchibald:
I think that should fix everything the review asked for... and actually, respond(somehow missed that). I've left the review open just in case though.

itsthejoker:
@TheLonelyGhost I'm fine with this; do you have any suggestions?

tor/core/user_interaction.py
```diff
@@ -328,3 +330,19 @@ def process_wrong_post_location(post):
             'Something went wrong with asking about a misplaced post; '
             'ignoring.'
         )
+
+def process_message(message: praw.models.Message, config):
+    author = message.author
+    username = author.name
```
itsthejoker: @TheLonelyGhost this will break if we receive a message without an author (e.g. anything from the Reddit platform itself).

TheLonelyGhost: Crap. You're right. I'll address that ASAP

PR #151, opened 2019-02-09
Noop bot run

TheLonelyGhost:
Adds in basic hooks for the bot to be started in noop mode (pronounced "no op" for "no operation"), where it only configures the bare essentials and then, instead of its usual bot duties, just waits until you CTRL-C to quit.

This is useful for testing changes to infrastructure so we can use the latest version of the code base, as opposed to the convention of using whatever is on the `CLOUD-DEPLOYMENT-TEST` branch. :sweat_smile:


TheLonelyGhost:
I know how much you love `argparse`, @itsthejoker, but I wasn't about to add a dependency for something simple that could've been solved by the standard library. 🤐

PR #153, opened 2019-02-11
First run for poetry and travis ci integration

itsthejoker:
This is for discussion and to force a Travis CI build to run. It is not suitable for merging at this point.


TheLonelyGhost:
@itsthejoker Is this suitable for review/merging now?

pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+repository = "https://github.com/grafeasgroup/tor"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+sh = "^1.12"
+tor_core = {git = "https://github.com/grafeasgroup/tor_core"}
+pytest = {version = "^4.2",optional = true}
+pytest-cov = {version = "^2.6",optional = true}
+codacy-coverage = {version = "^1.3",optional = true}
```
TheLonelyGhost: This probably shouldn't be part of the regular dependencies. We can manually pip install it on the Travis-CI run, since that's the only time we would need it installed.

pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+license = "MIT"
+repository = "https://github.com/grafeasgroup/tor"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+sh = "^1.12"
+tor_core = {git = "https://github.com/grafeasgroup/tor_core"}
+pytest = {version = "^4.2",optional = true}
+pytest-cov = {version = "^2.6",optional = true}
```
TheLonelyGhost: PyTest and PyTest-Cov seem like dev dependencies too. Do we use them at runtime for operating the bot?

.travis.yml
```diff
@@ -5,22 +5,28 @@ dist: trusty
...
+    - python: 3.7.2
     - python: nightly
 
+before_install:
+  - pip install poetry
+
 install:
-  - pip install --process-dependency-links .[dev]
-  - pip install codacy-coverage
+  - poetry install --no-dev --extras ci
```
TheLonelyGhost: We want to install it without the dev dependencies? Wouldn't we want to install it with both dev and CI deps so we'd be able to run automated tests and stuff?

pyproject.toml
itsthejoker: We don't need them at runtime, just for development. Here's how Poetry's installs break down:

* regular dependency: will install regardless.
* regular dependency, marked optional: will install if run from `poetry install`, will be absent from setup.py.
* dev dependency: will install if run from `poetry install`, has flag to disable and will be absent from setup.py.
* extras dependency: references regular dependencies and form the optional install versions.

For example, what I've got here renders out to this:

```python
install_requires = \
['sh>=1.12,<2.0', 'tor_core']

extras_require = \
{'ci': ['pytest>=4.2,<5.0',
        'pytest-cov>=2.6,<3.0',
        'codacy-coverage>=1.3,<2.0']}
```

pyproject.toml
itsthejoker: That defeats the purpose of having all the dependencies in one place, though 😛 
Either way, it will be only be installed when the extras `ci` package is installed.

pyproject.toml
itsthejoker: and actually, I also made a mistake about how that works. Lemme see about moving pytest under dev dependencies.

.travis.yml
```diff
@@ -5,22 +5,28 @@ dist: trusty
...
+  # - poetry add mypy
+  # - poetry add black --allow-prereleases
 
 script:
   - pytest --cov=tor
```
TheLonelyGhost: Shouldn't this be changed to the following?

```suggestion
  - poetry run pytest --cov=tor
```

TheLonelyGhost:
Deferring poetry to when we need it. At the moment, it would take more changes to our Chef cookbook than I'm comfortable with to actually use this so I'm going to close this until we have a firm path forward there.

PR #154, opened 2019-02-15
[RFC] Habitat deploy

TheLonelyGhost:
This is the culmination of a bunch of research I've been doing for months, really intended for other purposes but it seemed like it was a good fit here too.

Using [Habitat](https://www.habitat.sh/), we're able to deterministically build (and deploy) versions of the ToR bots. Here's a quick rundown of the major benefits:

- Declares system dependencies (e.g., libxml for BeautifulSoup4) in a way similar to what virtualenv does for Python packages, and makes the path to containerization trivial
- Service management to scale up and down or change configs super easy (think what Kubernetes does for containers, but it doesn't care if it's a container or some process on a VM or whatever)
- Software updates by subscribing to a "stable", "unstable", etc. channels (automatic upgrades when new release in subscribed channel)
- Software releases with options of several types of rolling upgrades (and rollbacks)
- Results of health checks made available over HTTP (basically making the CherryPy part of the ToR bot redundant)

Long story short, this should give us a more wholesome deployment story. More info: https://grahamweldon.com/post/2018/08/what-is-habitat/

Please as any questions you want about how Habitat works, why it would be helpful, and criticizing anything you don't understand or think is poorly designed. That's what I'm hoping to get out of this RFC.


TheLonelyGhost:
Needs to be significantly reworked. RFC has been given plenty of time, sounds like there are no objections.

PR #155, opened 2019-03-19
RFC: Add Code of Merit, modified for GG

itsthejoker:
One of the most important aspects of our work, both as an organization and in the codebases, is to be accepting of different viewpoints which can further our mutual mission. This naturally leads us in pursuit of some guidelines to govern contributor behavior and act as a higher set of rules with which we should use to resolve conflict. 

While a large number of projects include the Contributor Covenant as a guideline for contributor behavior, there are concerning issues that lead to multiple interpretations of the document and a lack of defined boundaries -- if a code of conduct is open to interpretation, then it is not strict enough.

The below is a modified version of the Code of Merit, slightly updated with an addition from the American Librarian Association (ALA), whose ethics guidelines have been well-understood and accepted for the last eighty years.

This PR will remain open for at least one (1) week, closing on 2019-03-26 if no conversation appears. Otherwise it will close when majority of contributors who wish to weigh in agree.

Signed-off-by: Joe Kaufeld <joe@grafeas.org>


emmademontford:
I love it! 

Kind regards, 

Emma Abdalla

itsthejoker:
Seven days are up, merging Code of Merit.

PR #160, opened 2019-04-20
Add "Unclaim" text

Pf441:
The bot will tell people about "Unclaim" after their claim (at the same time as the "Done"). It should help new (and older) transcribers to unclaim posts.


TheLonelyGhost:
This is great! Thanks for the contribution!

PR #161, opened 2019-06-07
[HOTFIX] parameterize hardcoded usernames

TheLonelyGhost:


TheLonelyGhost:
Travis-CI is failing for known reasons. Unfortunately this is one of those situations that should work _in theory_ but the only way to test is to release it.

tor/main.py
```diff
@@ -5,6 +5,7 @@
 from tor_core.helpers import run_until_dead
 from tor_core.initialize import build_bot
 
+import tor
```
itsthejoker: Let's mix this import with the one below it

itsthejoker:
```python
from tor import __version__, __SELF_NAME__, __BOT_NAME__
```

tor/main.py
```diff
@@ -77,6 +78,11 @@ def main():
 
     build_bot(bot_name, __version__, full_name='u/ToR')
     config.perform_header_check = True
+
+    tor.__SELF_NAME__ = config.r.user.me.name
+    if tor.__SELF_NAME__ not in tor.__BOT_NAMES__:
+        tor.__BOT_NAMES__.append(tor.__SELF_NAME__)
```
itsthejoker: This is an excellent idea!

itsthejoker:
I'm mostly down for it -- one stylistic change and then I think it's good to push.

tor/main.py
TheLonelyGhost: The reason these are separate is because we _want_ to mutate global state of `__SELF_NAME__`. If we refer to it as `__SELF_NAME__` only, it's just like any other variable we're shadowing (redefining over the top of it, not affecting the original) whereas assigning it like this: `tor.__SELF_NAME__ = 'foo'`, it affects the reference of it instead. It's assigning a value to the attribute of a global item that isn't being mutated directly, but rather just one of its attributes' values.

Difficult to explain, but this is some magic where we _want_ to affect global state and we can't do it if we do what you're suggesting.

itsthejoker: I am mildly horrified that this is a thing, but I understand what you're explaining it. Rock on 👍

PR #164, opened 2019-07-01
documentation update and extend detecting transcriptions in the wrong place

itsthejoker:
Right now we can detect transcriptions when they've been put in the r/ToR thread, but not if they're done using the new reddit formatter. This PR extends the current functionality to include the WYSIWYG editor mangling our footer.

Signed-off-by: Joe Kaufeld <joe.kaufeld@gmail.com>


TheLonelyGhost:
(currently holding off on merging until I have time to monitor the bots)

PR #166, opened 2019-08-15
Update flair.py

Pf441:
Fix transcription count ranking (and add Ruby).

### DIFF
--- a/tor/helpers/flair.py
+++ b/tor/helpers/flair.py
@@ -36,15 +36,17 @@ def flair_post(post, text):
 
 
 def _get_flair_css(transcription_count):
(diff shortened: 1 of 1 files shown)


itsthejoker:
We want to keep the `>=` bits because otherwise flair won't roll over until 501, 1001, etc. That's in line with how we originally did it, but that was kind of a mistake. The ruby part is fine, though!

Pf441:
When I saw the update on the Discord, I thought that I should change again levels in this request (and it's more logic to put them at round numbers 8-) ).

itsthejoker:
LGTM

PR #174, opened 2019-12-02
Lessens dependence on English language

TheLonelyGhost:
Logic was discussed in modchat, but comparison of the Comment's subreddit will determine if it was a username mention or a directive to the bot (if not in the ToR subreddit, must be a mention).

Also includes some other cleanup stuff, such as relying on redis instead of python for if member of the `blacklist` set.

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -157,19 +157,20 @@ def check_inbox(cfg):
     for item in reversed(list(cfg.r.inbox.unread(limit=None))):
         # Very rarely we may actually get a message from Reddit itself.
         # In this case, there will be no author attribute.
-        if item.author is None:
+        author_name = item.author.name if item.author else None
+        if author_name is None:
             send_to_modchat(
                 f'We received a message without an author -- '
                 f'*{item.subject}*:\n{item.body}', cfg
             )
             item.mark_read()
 
-        elif item.author.name == 'transcribot':
+        elif author_name == 'transcribot':
             item.mark_read()
 
-        elif item.author.name in cfg.redis.smembers('blacklist'):
+        elif author_name in cfg.redis.smembers('blacklist'):
             logging.info(
-                f'Skipping inbox item from {item.author.name} who is on the '
+                f'Skipping inbox item from {author_name!r} who is on the '
                 f'blacklist '
             )
             item.mark_read()
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -155,24 +155,22 @@ def check_inbox(cfg):
                 f'We received a message without an author -- '
                 f'*{item.subject}*:\n{item.body}', cfg
             )
-            item.mark_read()
 
         elif author_name == 'transcribot':
-            item.mark_read()
+            # bot responses shouldn't trigger workflows in other bots
+            logging.info('Skipping response from our OCR bot')
 
         elif cfg.redis.sismember('blacklist', author_name):
             logging.info(
                 f'Skipping inbox item from {author_name!r} who is on the '
                 f'blacklist'
             )
-            item.mark_read()
 
         elif isinstance(item, RedditComment):
             if str(item.subreddit.name).casefold() == str(cfg.tor.name).casefold():
                 # This should make it consistent behavior even if we're testing
                 # See: https://docs.python.org/3/library/stdtypes.html#str.casefold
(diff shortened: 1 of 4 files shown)


tor/core/inbox.py
```diff
@@ -157,50 +148,55 @@ def check_inbox(cfg):
...
-        elif item.author.name == 'transcribot':
+        elif author_name == 'transcribot':
             item.mark_read()
 
-        elif item.author.name in cfg.redis.smembers('blacklist'):
+        elif cfg.redis.sismember('blacklist', author_name):
             logging.info(
-                f'Skipping inbox item from {item.author.name} who is on the '
+                f'Skipping inbox item from {author_name!r} who is on the '
                 f'blacklist '
```
itsthejoker: Extra space at the end of the line

tor/core/inbox.py
```diff
@@ -157,50 +148,55 @@ def check_inbox(cfg):
...
 
-        elif item.author.name == 'transcribot':
+        elif author_name == 'transcribot':
             item.mark_read()
 
-        elif item.author.name in cfg.redis.smembers('blacklist'):
+        elif cfg.redis.sismember('blacklist', author_name):
             logging.info(
-                f'Skipping inbox item from {item.author.name} who is on the '
+                f'Skipping inbox item from {author_name!r} who is on the '
```
itsthejoker: Not sure what `author_name!r` is -- is that some formatting that I'm unfamiliar with or a typo?

tor/core/inbox.py
```diff
@@ -157,50 +148,55 @@ def check_inbox(cfg):
...
+        if author_name is None:
             send_to_modchat(
                 f'We received a message without an author -- '
                 f'*{item.subject}*:\n{item.body}', cfg
             )
             item.mark_read()
 
-        elif item.author.name == 'transcribot':
+        elif author_name == 'transcribot':
             item.mark_read()
```
itsthejoker: We can return here and just skip the rest of the processing since we know that we aren't going to do anything else to this comment.

TheLonelyGhost: There's an argument that we could just add the transcribot user to the blacklist and not even need this at all too.

Realistically, the `if`-`elif`-`else` tree makes it so no other condition is evaluated already. Returning here (though I think you mean `continue` since it's a for-loop) would be redundant.

tor/core/inbox.py
TheLonelyGhost: For those following along, it's a conversion operator in the formatting syntax spec for python: https://docs.python.org/3/library/string.html#format-string-syntax

tor/core/inbox.py
arfie: You could move `item.mark_read()` to the end of the function, since every branch calls it anyway. Then replace this one with just `pass`.

I also like the idea of blacklisting transcribot but that would probably need to be documented somewhere.

tor/core/inbox.py
```diff
@@ -76,7 +76,6 @@ def process_reply(reply, cfg):
     try:
         if any([regex.search(reply.body) for regex in MOD_SUPPORT_PHRASES]):
             process_mod_intervention(reply, cfg)
-            reply.mark_read()
             return
```
arfie: Similar here, instead of `return`ing every time, it could be a lot cleaner using if-elif?

tor/core/inbox.py
```diff
@@ -157,50 +148,56 @@ def check_inbox(cfg):
...
-        elif item.subject in ('comment reply', 'post reply'):
-            process_reply(item, cfg)
 
-        elif item.subject[0] == '!':
-            # Handle our special commands
-            process_command(item, cfg)
-            item.mark_read()
-            continue
+        elif isinstance(item, RedditComment):
+            if str(item.subreddit.name).casefold() == str(cfg.tor.name).casefold():
```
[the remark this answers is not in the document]
TheLonelyGhost: Ooh, that's a good call. I feel like this is a rabbit hole we could (and should) go down as we experience issues though so I'm not going to dig too deep in this iteration. Just enough to cover the basics. :)

tor/core/inbox.py
TheLonelyGhost: Meh, this was just a code style thing to say "before anything else is processed, do any of these apply?" sort of block. It doesn't really matter to me all that much though. 👍

PR #178, opened 2020-02-22
Blossom Functionality

itsthejoker:
There are a lot of changes in here, and as can pretty much be expected, not a lot of it is tested. For actually deploying this, I recommend setting the sub to private over a 2 hour window to verify that everything is working appropriately and any emergency fixes can be deployed.

Shitty, but that's what we get for doing large system changes without proper tests 🤷‍♂ 

If you see something wrong, let me know ASAP and I can resolve. This PR should move complete post and volunteer handling over to blossom.


TheLonelyGhost:
Merge conflicts are probably best resolved by using #180 instead. I'll leave this branch around to grab any other changes that might have been missed when re-coding this changeset.

PR #186, opened 2020-05-13
Refactor to allow Blossom Integration

MaxVanDeursen:
This is a draft PR which in the end should contain all code necessary to completely integrate `tor` with `Blossom`. This will be done in an iterative fashion, with PRs on the `blossom-refactor` branch to allow for structured reviewing along the way.

TODO's (apart from integrating all functionality with Blossom):
- [ ] Remove all TODOs from code
- [x] Create methods for common occurring code
- [ ] Make sure all actions are logged.


MaxVanDeursen:
> Wasn't there a ping function as well? It's probably not necessary in the long run, but the thought behind that was a quick check to make sure that blossom was responding when u/ToR starts up. If it's not needed then we can just drop the endpoint in blossom tbh.

I would definitely keep the quick pinging available for Blossom and you are right that there was such method within the Wrapper, but what I would like to do is start with this wrapper and then add methods along the way, converting `tor` to use Blossom step by step. So this would mean that some point in this rewriting, the `ping` method will be added to the wrapper.

tor/helpers/flair.py
```diff
@@ -56,72 +55,29 @@ def _get_flair_css(transcription_count: int) -> str:
...
-        user_flair = f'{new_count} Γ'
-        # add in that special flair bit back in to keep their flair intact
-        user_flair += additional_flair_text
-
-        cfg.tor.flair.set(post.author, text=user_flair, css_class=flair_css)
-        log.info(f'Setting flair for {post.author}')
-
-        post_author.update('transcriptions', current_transcription_count + 1)
-        post_author.save()
+    flair_postfix = " - Beta Tester"
```
MaxVanDeursen, line 65: Do we still want this to be the postfix after deploying it? IIRC this would be us going out of beta.

itsthejoker (dismissed):
As always, great work :+1:

Issue #187, opened 2020-05-16
Rephrase claim.already_claimed


MaxVanDeursen:
Currently the user receives this message whenever the post is already claimed, albeit by themselves or another user. We should probably reflect this in the `already_claimed` message.

TimJentzsch:
This is probably the same as #115, right?

PR #188, opened 2020-05-16
Add Blossom integration for the "Claim" message processing

MaxVanDeursen:
Add integration for the "claim" message processing. Note that this removes the need for the `already_completed` message, as this falls under the error message of `already_claimed`. Please also note #187 on this topic.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -5,9 +5,10 @@ from praw.exceptions import APIException, ClientException  # type: ignore
 from praw.models import Comment, Message  # type: ignore
 
 from tor import __BOT_NAMES__
+from tor.core.blossom_wrapper import BlossomStatus
 from tor.core.config import Config
 from tor.core.helpers import (_, clean_id, get_parent_post_id, get_wiki_page,
-                              reports, send_to_modchat)
+                              reports, send_reddit_reply, send_to_modchat)
 from tor.core.users import User
 from tor.core.validation import verified_posted_transcript
 from tor.helpers.flair import flair, flair_post, update_user_flair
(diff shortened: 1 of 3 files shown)


itsthejoker:
I'm not sure I understand why we want to remove `already_completed`, as that's supposed to trigger when someone `done`'s a completed post. It doesn't really make sense to submit `done` and get "Someone has already claimed this" in response...

tor/core/user_interaction.py
```diff
@@ -76,60 +77,37 @@ def process_coc(post: Comment, cfg: Config) -> None:
...
     """
-    Handles comment replies containing the word 'claim' and routes
-    based on a basic decision tree.
+    Process a claim request.
 
-    :param post: The Comment object containing the claim.
-    :param cfg: the global config dict.
-    :return: None.
+    This function sends a reply depending on the response from Blossom and
+    creates an user when this is the first time a user uses the bot.
```
itsthejoker, line 83: Need to add a TODO item for handling `first_time` as well.

MaxVanDeursen:
>I'm not sure I understand why we want to remove already_completed, as that's supposed to trigger when someone done's a completed post. It doesn't really make sense to submit done and get "Someone has already claimed this" in response...

I think you are mistaken, the code falls under the `claim` category and is only used in the case where a `claim` is used on a completed post. However in this case it would suffice to say that the post is already claimed, which is what I have done.

tor/core/user_interaction.py
MaxVanDeursen: In the original code, `first_time` was used to determine the success message, which is done in line 100 of the new code.

itsthejoker: Shit, you're right. Sorry!

MaxVanDeursen:
@itsthejoker Does this mean that this PR can be merged?

PR #192, opened 2020-05-21
fix type error for blossom_wrapper

itsthejoker:
Adds `black` as a development dependency so that we can use it.

The primary fix here is that it solves an error with resolving types in the dataclass, namely allowing None as a valid option since it's initialized that way.

Signed-off-by: Joe Kaufeld <joe.kaufeld@gmail.com>

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,5 +1,13 @@
 [[package]]
 category = "dev"
+description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
+name = "appdirs"
+optional = false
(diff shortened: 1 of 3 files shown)


MaxVanDeursen:
Can we maybe take this a step further and add the complete `flake8` configuration from Blossom into here, alongside the GitHub static analysis action from Blossom?

itsthejoker:
I don't have an issue with that. Let's merge this and then I can make a new PR for testing configurations.

MaxVanDeursen:
Sure!

PR #193, opened 2020-05-29
Add Blossom integration for the "Done" message processing

MaxVanDeursen:
In this PR the integration is done for the "Done" procedure. Note that the methods used within the old workflow are also refactored within this PR.


MaxVanDeursen:
@itsthejoker After a long hiatus, I have incorporated your feedback

PR #194, opened 2020-06-02
Added note about footer on Reddit Mobile

Altrissa:
Users cannot copy/paste the footer using reddit mobile. Added a note in the footer comment about it.

### DIFF
--- a/tor/strings/en_US.yml
+++ b/tor/strings/en_US.yml
@@ -33,6 +33,8 @@ posts:
     Footer
     ---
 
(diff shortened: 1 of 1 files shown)


MaxVanDeursen:
I would recommend moving the note to one line below

TheLonelyGhost:
Merged because it looked fine to me. Now I see the previous comment and I see it was premature. Is it enough of a sticking point, @MaxVanDeursen, that we should make that change in an additional PR?

PR #198, opened 2020-07-24
Remove further Redis usage from the code

MaxVanDeursen:
In this PR, the following Redis calls are removed, and are replaced with Blossom integration:

- Redis calls for accepting the CoC
- Redis calls for Submission creation
- Redis calls for Blacklist checking

There are two things which require discussion:
- I would think that for adding people to the blacklist, Blossom should now be used rather than doing this through `tor`. Hence, I removed the "adding users to the blacklist" command from `tor`'s functionality
- Right now, we use redis as a queue for the `ocr` bot to create transcriptions. I think that we ultimately would want to do this differently, either through Blossom or the `ocr` bot checking for new ToR posts himself.


itsthejoker:
The last time we talked about it, we had decided to leave redis functionality specifically for informing u/transcribot of new posts... but you have a valid point that it probably can and should be moved to Blossom. I'll add a new issue.

tor/core/blossom_wrapper.py
```diff
@@ -154,6 +154,32 @@ def get_user(self, username: str) -> BlossomResponse:
         else:
             return BlossomResponse(status=BlossomStatus.not_found)
 
+    def accept_coc(self, user_id: str) -> BlossomResponse:
+        """Let the user accept the Code of Conduct."""
+        response = self.get(f"/volunteer/{user_id}/accept_coc")
```
itsthejoker, line 159: needs to move to `/volunteer/accept_coc/?username=theguy`

tor/core/inbox.py
```diff
@@ -154,33 +155,31 @@ def check_inbox(cfg: Config) -> None:
...
-                process_message(item, cfg)
-
+            log.info("Skipping response from our OCR bot")
         else:
-            # We don't know what the heck this is, so just send it onto
-            # slack for manual triage.
-            forward_to_slack(item, cfg)
-
+            user_response = cfg.blossom.get_user(username=author_name)
+            if user_response.status != BlossomStatus.not_found and user_response.data["blacklisted"]:
```
itsthejoker, line 166: You shouldn't have to ask for the answer now that the blacklist has been fully implemented; you should get a 423 Locked response code when attempting an action with a blacklisted user.

tor/core/posts.py
```diff
@@ -63,54 +70,45 @@ def has_enough_upvotes(post: PostSummary, cfg: Config) -> bool:
...
-    if subreddit not in cfg.upvote_filter_subs:
-        # Must not be a sub which has a minimum threshold
-        return True
-
-    if upvotes >= cfg.upvote_filter_subs[subreddit]:
-        return True
-
-    return False
+    # If the subreddit is not in the upvote filter, this would mean no threshold.
+    return upvotes >= cfg.upvote_filter_subs.get(subreddit, float("-inf"))
```
itsthejoker, line 74: I think negative infinity may be a little overkill but it works :smile:

tor/core/user_interaction.py
```diff
@@ -17,61 +16,54 @@
...
-    # Have they already been added? If 0, then just act like they said `claim`
-    # instead. If they're actually new, then send a message to slack.
-    if result == 1:
+    """Process the acceptation of the CoC by the specified user."""
+    username = post.author. name
+    user_response = cfg.blossom.get_user(username=username)
+    if user_response.status == BlossomStatus.ok:
+        # The status codes are not checked because they are already caught by
+        # getting the user.
+        cfg.blossom.accept_coc(user_id=user_response.data.id)
```
itsthejoker, line 48: switch to username

tor/strings/en_US.yml
```diff
@@ -99,6 +99,8 @@ responses:
     claimed_other_user: |
       Sorry, this post is claimed by someone else, so you cannot unclaim it!
   general:
+    blacklisted: |
```
itsthejoker, line 102: Historically we've built it so that they are just ignored by the bots... I think that unless we have a good reason to inform people of this fact, we should just continue ignoring them.

MaxVanDeursen: You're right, we could also do this. That being said, wouldn't it be easier to just make the respective user aware of the fact?

MaxVanDeursen:
@itsthejoker Would you mind reviewing this once again?

tor/strings/en_US.yml
itsthejoker: Blacklisting should be reserved exclusively for people who are abusing the platform, and it's not an automatic process -- it's one that we trigger manually. I perceive that we should follow the reddit shadowban system -- let them scream into the void in the hopes that something answers until they go away. We will have already made our intentions clear long before blacklisting becomes an option.

tor/helpers/reddit_ids.py
```diff
@@ -3,46 +3,10 @@
...
-    :param cfg: the global config object.
-    :return: True if the ID is successfully inserted into the set; False if
-        it's already there.
-    """
-
-    # if we get back a True, then we return True because the post was submitted
-    # successfully and it's good to work on. If the insert fails, then we
-    # want to return a False because we cannot work on that post again.
-    return add_complete_post_id(post_id, cfg)
+    # TODO: This check requires a check of existence in Blossom based on the "url"
```
itsthejoker, line 7: I'll work on this today.

itsthejoker (approved):

PR #206, opened 2021-03-24
Patch breaking bug for private subreddits

itsthejoker:
Wraps the call to load the target submission with a check to get a 403 forbidden, which is what Reddit returns through the API if you try and access a specific ID behind a private subreddit. This new behavior will fall through to the author check and, if successful, will not trigger the slack ping about a removed post.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tor"
-version = "4.2.2"
+version = "4.2.3"
 description = "A bot that handles moderating and scoring in /r/TranscribersOfReddit"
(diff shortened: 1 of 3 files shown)


TheLonelyGhost (changes requested):
Yep, looks pretty reasonable. We'll address the mypy failure in another iteration.

Only remaining problems to bump the version in `tor/__init__.py` and `pyproject.toml` and adding to the changelog

TheLonelyGhost (approved):
Looks good! Don't forget to tag the merge commit when it gets merged to `main`. It'll trigger the thing to package it as a wheel.

PR #208, opened 2021-04-10
Add warning about "I agree"

Pf441:
We have had a lot of new volunteers writing "I agree" or "accept" instead of "I accept", which is a problem since their claim won't be processed. I added an explicit warning about that in the message of the bot introducing the Code of Conduct.


itsthejoker (approved):
:tada:

PR #213, opened 2021-06-06
Honeycomb support

TheLonelyGhost:


MaxVanDeursen:
Looks sweet!

MaxVanDeursen (approved):

TheLonelyGhost:
Some fun with getting the annotations in `tor/core/config.py` to actually play nicely with mypy, but once I figure that out it should have green checkmarks all the way.

itsthejoker:
What do we need to do for deploying this?

TheLonelyGhost:
Needs some TLC to make mypy not fail spectacularly on upgrading to python 3.9.

I don't want to rip out the types again, but I'm not yet sure what the problem is that it has with them.

TheLonelyGhost:
Okay, looks like it may just require updating mypy

https://github.com/python/mypy/issues/9761

TheLonelyGhost:
Upgrading mypy to >= v0.800 is what fixed the python 3.9 support. Now we're on v0.902 of it, so I hope python 3.10 doesn't introduce any unexpected problems like the last "minor" upgrade.

scripts/clear_submissions_from_queue.py
```diff
@@ -1,13 +1,14 @@
 from datetime import datetime, timedelta
 
 from praw import Reddit
-from praw.exceptions import RedditAPIException
+# from praw.exceptions import RedditAPIException
```
MaxVanDeursen, line 4: Better not to comment code

MaxVanDeursen (dismissed):

scripts/clear_submissions_from_queue.py
itsthejoker: This is a throwaway script that's supposed to be modified in place depending on whatever we need -- if this is the only issue you have with the PR, then I would consider it good to go

TheLonelyGhost: ^^^ This.

Main reason I commented it out here instead of deleting is because some seemingly important try-except code was commented out later in the file, where it handles this particular exception. Rather than delete the commented code in its entirety, this file seems more like a living workspace just in case we need it. Might even be able to delete it if we rework it into one of the commands in the bot. 🤷

itsthejoker (approved):

PR #215, opened 2021-06-07
Send full transcription URLs to Blossom

itsthejoker:
Closes https://github.com/GrafeasGroup/blossom/issues/135.

Right now, we're sending a complete URL to Blossom when we're creating a new submission (https://github.com/GrafeasGroup/tor/blob/main/tor/core/posts.py#L130) but we don't format the URL when we send the transcription. This PR resolves this issue.


MaxVanDeursen (approved):

TimJentzsch:
While this fixes the links to be clickable, I'm not sure if it fixes the whole issue. These particular links are direct links to the transcription on the partner sub. And from what I've seen, some of them got deleted from the spam filter. This would make reviewing the transcription hard/impossible.

itsthejoker:
@TimJentzsch I'm going to merge this because it fixes part of the issue; the other part of the issue, if we change it, needs to happen in Blossom.

PR #218, opened 2021-06-08
recover cleanly from top-level deleted comments

itsthejoker:


tor/core/validation.py
```diff
@@ -26,6 +26,11 @@ def get_transcription(
     linked_submission = cfg.r.submission(url=submission_url)
     linked_submission.comments.replace_more(limit=0)
     for top_level_comment in linked_submission.comments.list():
+        try:
+            # Watch out for top level deleted comments
+            top_level_comment.author.name
+        except AttributeError:
+            continue
```
TimJentzsch, line 33: I think this would be better be done in an `if` check.

TimJentzsch (changes requested):

TimJentzsch (approved):

PR #228, opened 2021-07-18
Add debug command

itsthejoker:
This PR is fairly large for the actual number of material changes in it; most of it is `black` and various cleanups that I stumbled over while implementing this command. Full list of changes:

* update donor comment
* move all common string phrases to `tor/core/__init__.py` to allow importing from more than one place
* update `!override` to work on any source comment that is valid for claim or done, like `deno`
* don't attempt to send the "hey I don't respond to DMs" message to an account that doesn't exist (for the messages from reddit as a platform)
* add a `!debug` command that just prints out the Blossom object for the current submission. Should make the next step in debugging immediately obvious when working with a troublesome submission.


TimJentzsch (approved):
Looks good!

PR #230, opened 2021-07-18
closes #227: fix override breaking if no transcription is found

### LINKED ISSUE #227: Fix `override` command not working
The `override` command to award users gamma in case something goes wrong is not working in some instances.

My initial guess is that it *requires* to find a transcription of the user, but sometimes that's not available due to Reddit being stupid.
In these cases it would probably have to create a dummy transcription for the gamma to be awarded.

Examples:
- https://www.reddit.com/r/TranscribersOfReddit/comments/ohme08/peanutbutterisoneword_image_not_my_pic_but_this/
- https://www.reddit.com/r/TranscribersOfReddit/comments/ogxr4h/feminism_image_call_for_contributions_a_global/h4mx88x/

itsthejoker:
When trying to perform an override, the current process hinges on whether or not a transcription is found. If the transcription is removed (or doesn't exist) then the override will fail. This PR moves the actual "done" logic to a separate check to see if we find the transcription OR an override has been triggered.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -180,7 +180,7 @@ def process_done(
 
     if transcription or override:
         # because we can enter this state with or without a transcription, it
-        # makes sense to have this is a separate block.
+        # makes sense to have this as a separate block.
         done_response = cfg.blossom.done(
             blossom_submission["id"], user.name, override
         )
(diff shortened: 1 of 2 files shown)


tor/core/user_interaction.py
```diff
@@ -178,6 +178,9 @@ def process_done(
             not is_visible
         )
 
+    if transcription or override:
+        # because we can enter this state with or without a transcription, it
+        # makes sense to have this is a separate block.
```
TimJentzsch, line 183: to have this is a?

TimJentzsch (approved):
Looks good, just one typo in a comment :)

PR #231, opened 2021-07-18
Fix default response for claims to better identify iffy responses from Blossom

itsthejoker:
Found a bug in blossom-wrapper where something that was already claimed was being marked as already completed, and that has already been fixed in the wrapper. This PR makes the correct response return (instead of defaulting to "this has already been claimed") so that we have a better idea of when something explodes instead of it giving the wrong response.

### DIFF
--- a/tor/core/user_interaction.py
+++ b/tor/core/user_interaction.py
@@ -112,8 +112,11 @@ def process_claim(
     elif response.status == BlossomStatus.blacklisted:
         message = i18n["responses"]["general"]["blacklisted"]
 
-    else:
+    elif response.status == BlossomStatus.already_claimed:
         message = i18n["responses"]["claim"]["already_claimed"]
+
+    else:
+        message = i18n["responses"]["general"]["oops"]
 
     return message, return_flair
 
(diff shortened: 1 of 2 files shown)


tor/core/user_interaction.py
```diff
@@ -112,9 +112,12 @@ def process_claim(
     elif response.status == BlossomStatus.blacklisted:
         message = i18n["responses"]["general"]["blacklisted"]
 
-    else:
+    elif response.status == BlossomStatus.already_claimed:
         message = i18n["responses"]["claim"]["already_claimed"]
 
+    else:
+        message = i18n["responses"]["general"]["oops"]
+
```
TimJentzsch, line 120: Does it give us a Slack ping in this case? If not, I think it should

TimJentzsch (approved):

tor/core/user_interaction.py
itsthejoker: It doesn't give us a ping, but this would be a great thing to add. Maybe a bugsnag hit instead? Either way, it's outside the scope of this.

PR #235, opened 2021-07-23
Add more valid words for commands

itsthejoker:
Since we've been seeing some additional new misspellings come through, seems like a reasonable idea to add them to the bot since it's pretty clear what they're trying to attempt.

### DIFF
--- a/tor/core/__init__.py
+++ b/tor/core/__init__.py
@@ -56,6 +56,6 @@ MOD_SUPPORT_PHRASES = [
     # re.compile('(?:good|bad) bot', re.IGNORECASE),
 ]
 
-CLAIM_PHRASES = ["claim", "dibs"]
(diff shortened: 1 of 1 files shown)


tor/core/__init__.py
```diff
@@ -57,5 +57,5 @@ def __get__(self, obj, _type=None):
 ]
 
 CLAIM_PHRASES = ["claim", "dibs"]
```
TimJentzsch, line 59: I think this would be a good time to also add some common `claim` mistypes, such as `clam` and `caim` and `clai`.

TimJentzsch (approved):

tor/core/__init__.py
itsthejoker: yep, good call

PR #238, opened 2021-07-24
Automatically detect common formatting issues in transcriptions

### LINKED ISSUE #234: RFC: Automatically detect common formatting issues
A lot of formatting issues are very common with new transcribers. For example, a lot of people use the Fancy Pants editor instead of Markdown Mode.

Some of these issues could be detected automatically. For example, the Fancy Pants issue can be detected by looking for an escaped footer.

When a transcription with formatting issues is detected, the bot should reject the `done` and present the user a list with the detected issues and instructions on how to fix them. It should also offer a away to get a human moderator to help (e.g. a `help` command).

The goal of this would be to avoid butchered transcriptions being uncaught and to reduce repetitive workload for the welcome mods.

TimJentzsch:
Closes #234 

## Description

This adds checks to automatically detect common formatting issues in transcriptions.
It currently checks for the following:

- Headers being bold instead of italic
- Missing horizontal separators
- Separators being misused as headings
- Malformed footers
- Fenced code blocks

When one or more issues are detected, the `done` is rejected and the users and mods get notified.

Originally I planned to add detection for Fancy Pants usage instead of Markdown Mode, but it seems like Reddit fucked up their editor so that you can't properly copy paste stuff anymore. So maybe we will see less of those anyway.

This also adds `help` as a keyword to notify the mods on Slack and adds a message on Reddit to tell the user that help is on the way.


itsthejoker (changes requested):
This looks amazing, but there are still some imports left for `tor.core.validation` -- both mypy and the tests are failing :/

itsthejoker (approved):

itsthejoker (approved):

Issue #237, opened 2021-07-24
Fix u/transcribersofreddit spamming comments on YT posts


TimJentzsch:
When u/transcribersofreddit finds subtitles on a YT post, it will spam messages on the post on the partner sub:
https://www.reddit.com/r/aaaaaaacccccccce/comments/oqgl2b/garlic_bread_went_to_space_before_the/?utm_medium=android_app&utm_source=share

This needs to be fixed ASAP

PR #239, opened 2021-07-24
Remove youtube transcript auto-detection feature

itsthejoker:
This feature is rarely used. It exploded on a partner subreddit this morning, and it seems like more trouble than it's worth for this feature, so this PR removes the ability for ToR to attempt to automatically detect human-written transcriptions on youtube videos.

By removing this, it also limits ToR to only posting in our subreddit; this functionality is the only one that allows ToR to venture beyond our walls.


itsthejoker:
Closes #237

TimJentzsch (approved):

PR #243, opened 2021-07-25
Handle double claiming and mass claims

### LINKED ISSUE #187: Rephrase claim.already_claimed

### LINKED ISSUE #220: Prevent mass-claiming of posts
Currently, when someone claims multiple posts at once, they'll receive a warning from the moderators. Instead, the bot should prevent this in the first place. 
I think a limit of one claim at a time for new users and maybe two at a time for 100+ gamma would be sensible (in case of someone doing a long video transcription and needing a break with some text posts).
If a claim over this limit is attempted, a message like this should be sent:
```
Whoa, slow down there! It seems like you already claimed the following post(s):
- <link to post 1>
- <link to post 2>
- ...
Please finish them before claiming a new one. If you don't want to transcribe these posts anymore, please reply `unclaim` as a top level comment on these posts.
```
And the claim should be blocked.

TimJentzsch:
Relevant issues: Closes #220, closes #187.

This properly handles the user trying to claim a post twice. Instead of saming that "someone" already claimed the post, it will tell the user that they already got the post. If another person claimed it, the bot will tell you who it was (without pinging them).

Additionally, this adds handling for the mass-claim prevention. When a user claimed to many posts, the bot will give them a list of submissions that they still need to complete.

⚠️  This change should be deployed immidiately **before** deploying GrafeasGroup/blossom#183 and GrafeasGroup/blossom#185. ⚠️ 
These PRs contain the API changes necessary for these responses.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ bugsnag = "^3.6"
 requests = "^2.22"
 slackclient = "^1"
 PyYaml = "^5.1"
-blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master", rev = "e8ed76b" }
+blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master" }
 python-dotenv = "^0.14.0"
 praw = "^7.1.0"
 toml = "^0.10.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ bugsnag = "^3.6"
 requests = "^2.22"
 slackclient = "^1"
 PyYaml = "^5.1"
-blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master" }
+blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master", rev = "e8ed76b" }
 python-dotenv = "^0.14.0"
 praw = "^7.1.0"
 toml = "^0.10.2"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ bugsnag = "^3.6"
 requests = "^2.22"
 slackclient = "^1"
 PyYaml = "^5.1"
-blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master" }
+blossom-wrapper = { git = "https://github.com/GrafeasGroup/blossom-wrapper.git", branch = "master", rev = "e8ed76b" }
```
itsthejoker, line 24: I'd actually leave this at master -- we have the ability to deploy everything at once, so it would make the most sense to have this stay the same.

TimJentzsch: My thought process was that this makes it harder to have multiple branches open while updating the wrapper and we have to always remember to update all the bots at once -- I think we are bound to forget one at some point.

But I can change it back if you don't like it

itsthejoker: My concern is that the existing workflow is:

- make a slight change to the wrapper
- deploy the bot that relies on it, have that bot install it
- if multiple bots need the changes at once, then it's just `bubbles deploy all`

I worry that if we set this to a specific hash, we'll forget about it and newer changes won't get merged in. If we're worried about forgetting, maybe it would be a good idea to just make `bubbles deploy` default to refreshing everything.

TimJentzsch: But if you are worried that we forget about it, wouldn't that mean that with the current workflow that bot would just break, because it's not adapted to the new wrapper?

TimJentzsch: I moved this to a separate issue, #244.

itsthejoker (approved):

PR #245, opened 2021-07-28
Fix variable naming and wrong link access on comment

TimJentzsch:
This fixes that a variable for a Reddit comment was named `post` and fixes a faulty attribute access to get a comment link.


itsthejoker (approved):
let's do ittt

PR #252, opened 2021-07-29
Fix tests not passing and formatting issues

### LINKED ISSUE #250: Fix code blocks in formatting issue suggestions
Unfortunately, Reddit markdown is super inconsistent and lists are generated differently in Old and New Reddit.
On Old Reddit, every succeeding line of a list item needs to start with four additional spaces. On New Reddit it's two spaces.

Old Reddit:
List item test with two to four spaces. It displays as code block when using four additional spaces.

New Reddit:
List item test with two to four spaces. It displays as code block with two additional spaces. After that, the code block starts with more spaces.

We will have to choose a different method entirely, maybe using separators.

### LINKED ISSUE #251: Add more transcriptions as tests for formatting validation
We should add a rather large set of transcriptions as test for the formatting validation, to make sure that they don't generate any false positives. We could either pull them directly from blossom, e.g. by providing a list of valid transcription IDs, but that would require the usage of an API key and potentially be slow. Or we provide a folder with many sample transcriptions downloaded from the API. That would be faster and easier, but add a lot of files to the bot.

TimJentzsch:
Closes #250, closes #251.

This PR addresses some issues that were left with the formatting validation.

- It turns out that you _can't_ make code blocks within list items and have them display correctly on both old and new Reddit. Instead, we now use horizontal separators between them. Hopefully this finally works.
- Some of the regex patterns used for the formatting validation had misleading names and no documentation. I fixed that and hopefully made it a lot clearer.
- A lot of tests were not passing anymore, partially due to the bad documentation mentioned above. I fixed that and added more tests to make sure that everything is working correctly.
- A test suit has been added that checks that all transcriptions provided in a folder pass without formatting issues. This contains a selection of real transcriptions form multiple volunteers and should allow us to be more confident that the formatting checks don't return false positives.


tor/strings/en_US.yml
```diff
@@ -201,29 +201,29 @@ formatting_issues:
...
     \    <header>\n
-    \   \n
-    \   ---\n
-    \   \n
-    \   <content>\n
-    \   \n
-    \   ---\n
-    \   \n
-    \   <footer>"
+    \    \n
```
itsthejoker, line 204: I removed a space from these because they were displaying as one space too far without them -- it's a weird quirk that I don't understand the meaning behind. Same with the extra space behind line 217.

tor/strings/en_US.yml
```diff
@@ -201,29 +201,29 @@ formatting_issues:
...
     \    ---\n
     \    (empty line)\n
 
 
     If you put it directly after a line of text, it will display as a heading, for example:
 
 
-    \  Heading\n
-    \ ---
+    \    Heading\n
```
itsthejoker, line 225: same here -- the top line needs to be indented by one more space than the rest

tor/strings/en_US.yml
```diff
@@ -245,17 +245,17 @@ formatting_issues:
   fenced_code_block: "**Fenced code block**: It looks like you used a fenced code block, i.e. you surrounded your code by three backticks, like this:
 
 
-    \  ```\n
-    \ int x = 0;\n
-    \ int y = x * 100;\n
-    \ ```
+    \    ```\n
```
itsthejoker, line 248: and here

itsthejoker (approved):

itsthejoker (changes requested):
Some issues with the yml that will cause display errors on reddit's side.

itsthejoker (approved):
let's do it!

PR #260, opened 2021-08-05
Added messages to update a user reaching a new rank.

Arnab1181412:


tor/core/user_interaction.py
```diff
@@ -256,6 +257,16 @@ def process_done(
                 f" successful."
             )
             message = done_messages["completed_transcript"]
+            
+            transcription_count = blossom_user["gamma"]
+            is_promoted = check_promotion(transcription_count)
```
TimJentzsch, line 262: I think we have to input `transcription_count + 1` here. The transcription count in the `blossom_user` object has been retrieved *before* the new transcription has been processed, so it will be one behind at this point.

tor/core/user_interaction.py
```diff
@@ -268,7 +279,7 @@ def process_done(
         elif done_response.status == BlossomStatus.blacklisted:
             message = i18n["responses"]["general"]["blacklisted"]
 
-    return message, return_flair
+    return message, return_flair, alt_promotion_txt
```
TimJentzsch, line 282: This will not work, as only two arguments are expected to be returned. Instead, we will have to modify the `message` variable to append the promotion text.
We could do something similar to the `alt_text_trigger` in line 270, e.g.

```py
message = f"{message}\n\n{alt_promotion_txt}"
```

tor/helpers/flair.py
```diff
@@ -59,6 +59,32 @@ def _get_flair_css(transcription_count: int) -> str:
...
+        return 'grafeas-diamond'
+    elif transcription_count == 500:
+        return 'grafeas-golden'
+    elif transcription_count == 250:
+        return 'grafeas-purple'
+    elif transcription_count == 100:
+        return 'grafeas-teal'
+    elif transcription_count == 50:
+        return 'grafeas-green'
+    elif transcription_count==1:
```
TimJentzsch, line 80: There seem to be some spaces missing here, you can run `black tor/helpers/flair.py` to run our automatic formatting tool on this.

tor/helpers/flair.py
```diff
@@ -59,6 +59,32 @@ def _get_flair_css(transcription_count: int) -> str:
         return 'grafeas'
 
 
+def check_promotion(transcription_count):
```
TimJentzsch, line 62: This is very similar to `_get_flair_css`, I wonder if we could combine them in some way?

tor/strings/en_US.yml
```diff
@@ -112,6 +113,30 @@ responses:
...
+      grafeas-golden: |
+        Yipee!! Your work is so golden.
+      grafeas-purple: |
+        Congratulations!! You have hit the 250 mark. Keep going!
+      grafeas-teal: |
+        Century!! You have just hit the magic number 100.
+      grafeas-green: |
+        Congratulations!! You have achieved your first rank.
+      grafeas-initiate: |
+        Hooray!! You just made your first transcription.
```
TimJentzsch, line 135: I think it would be good to have the number of transcriptions included in every message here, in case the volunteer is not familiar with the ranks.

Some of the numbers seem to be in the wrong message, e.g. Ruby is 2500 transcriptions and not 1000 (that's diamond). Here is a full list of the ranks and their transcription counts:

- 1 - 49: Initiates
- 50 - 99: Green
- 100 - 249: Teal
- 250 - 499: Purple
- 500 - 999: Gold 
- 1,000 - 2,499: Diamond
- 2,500 - 4,999: Ruby
- 5,000 - 9,999: Topaz
- 10,000+: Jade

TimJentzsch (changes requested):
Looking good already! I left some comments with suggestions for improvement :)

tor/core/user_interaction.py
Arnab1181412: Yes Sir I understood the logic

tor/helpers/flair.py
Arnab1181412: Sir I get what you are saying but I think if we combine them both then this function will definitely return something so I think we should'nt combine them both.

tor/helpers/flair.py
Arnab1181412: Sir I didn't get this point

TimJentzsch: It's just a small formatting issue, it should be `elif transcription_count == 1:`. We have an automatic formatting tool, called `black` in our Poetry environment, so if you used Poetry to set up your project you can use the `black tor/helpers/flair.py` command to fix the formatting for this file

TimJentzsch: Oh I see you already fixed it, nice!

TimJentzsch (approved):
This is looking good now! Thanks a lot for you help!

Arnab1181412:
Thank you sir

TimJentzsch:
Hello sir @Arnab1181412,
do you need help with fixing the formatting of the files?
These are the issues that our CI system detects:
Formatting issues in the PR detected by flake
You can run the `flake .` command locally to see it for yourself (in the Poetry environment of the project).
After that is fixed, we can merge the PR!

PR #261, opened 2021-08-06
Implement formatting check for long hyphen

fulminatingmoat:
Not too necessary since missing separator check will detect it as well


test/validation/test_formatting_validation.py
```diff
@@ -176,6 +177,22 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
     assert actual == expected
 
 
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("—-", True),
+        ("-—", True),
+        ("---------", False),
+        ("Word\n    int x = 1\nWord", False),
```
TimJentzsch, line 186: I would add `("---", False)` here as well, as this is the most commonly used separator string.

TimJentzsch (changes requested):
The implementation itself looks good, however we also need to define the response string for this formatting issue in `tor/strings/en_US.yml`. For examples you can take a look at the other formatting issues.

TimJentzsch (approved):

TimJentzsch (changes requested):
We still need to define the corresponding strings for the bot response.

TimJentzsch:
Here we still need to define the bot response for the volunteers, you can take https://github.com/GrafeasGroup/tor/blob/main/tor/strings/en_US.yml#L209-L319 as examples :)

tor/validation/formatting_validation.py
```diff
@@ -149,6 +153,24 @@ def check_for_fenced_code_block(transcription: str) -> Optional[FormattingIssue]
...
+    ---
+
+    Mobile devices may convert the separator to:
+    —- or -—
+
+    These don't display correctly on all devices
+    """
+    return (
+        FormattingIssue.AUTOGENERATED_LONG_HYPHEN
+        if AUTOGENERATED_LONG_HYPHEN_PATTERN.search(transcription) is not None
```
TheLonelyGhost, line 169: Why not drop the regex and do a simple `x in y` check?

```suggestion
        if "\u2014" in transcription  # \u2014 == em dash == "long hyphen"

```

test/validation/test_formatting_validation.py
```diff
@@ -242,6 +243,23 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
...
+    [
+        ("—-", True),
+        ("-—", True),
+        ("---------", False),
+        ("---"), False),
+        ("Word\n    int x = 1\nWord", False),
+    ],
+)
+def test_check_for_autogenerated_long_hyphen(test_input: str, should_match: bool) -> None:
+    """Test if autogenerated long hyphens are detected."""
```
TheLonelyGhost, line 257: The long hyphen isn't always auto-generated, so perhaps "accidental" would be a better descriptor than "autogenerated"?

Side note: if settings for "smart quotes" (and the like) are enabled on macOS, it automatically replaces `--` with `—` there too. Not just mobile. 😄

TheLonelyGhost (changes requested):

TheLonelyGhost:
No comments and no meaningful code updates in months. Closing this since it seems pretty stale to me. Feel free to reopen if the mood strikes to pick this up again!

PR #262, opened 2021-08-06
Add check for double-spaced and escaped line break

### LINKED ISSUE #249: Add formatting check for double-spaced and escaped line breaks
Unfortunately, Reddit is inconsistent with their Markdown parser and not all clients enforce double-spaced and escaped line breaks. Instead, a paragraph line break has to be used.

DO:
- Paragraph line break:
    ```
    This is a line
    
    This is another line
    ```

DON'T:
- Double-spaced line break (note the two spaces at the end of the first line):
    ```
    This is a line  
    This is another line
    ```
- Escaped line break:
    ```
    This is a line\
    This is another line
    ```

fulminatingmoat:
Fixes #249

### DIFF
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -9,6 +9,7 @@ from test.validation.helpers import (
 )
 from tor.validation.formatting_validation import (
     check_for_fenced_code_block,
+    check_for_incorrect_line_break,
     check_for_missing_separators,
     check_for_formatting_issues,
     check_for_heading_with_dashes,
@@ -173,6 +174,21 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
     """Test if fenced code blocks are detected."""
     actual = check_for_fenced_code_block(test_input)
     expected = FormattingIssue.FENCED_CODE_BLOCK if should_match else None
+    assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("This is a line  \nThis is another line", True),
+        ("This is a line\\\nThis is another line", True),
+        ("This is a line\n\nThis is another line", False),
+    ],
+)
+def test_check_for_incorrect_line_break(test_input: str, should_match: bool) -> None:
+    """Test if incorrect line breaks are detected."""
+    actual = check_for_incorrect_line_break(test_input)
+    expected = FormattingIssue.INCORRECT_LINE_BREAK if should_match else None
     assert actual == expected
 
 
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -183,6 +183,10 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
         ("This is a line  \nThis is another line", True),
         ("This is a line\\\nThis is another line", True),
         ("This is a line\n\nThis is another line", False),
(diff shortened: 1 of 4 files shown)


tor/validation/formatting_validation.py
```diff
@@ -42,6 +42,21 @@
...
+#
+# This is another line
+# DON'T:
+# Double-spaced line break (note the two spaces at the end of the first line):
+# This is a line
+# This is another line
+# Escaped line break:
+# This is a line\
+# This is another line
+INCORRECT_LINE_BREAK_PATTERN = re.compile(r"  \n|\\\n")
```
TimJentzsch, line 58: I think we will have to make this regex stricter to avoid false positives. It often happens that volunteers have extra spaces at the end of a line, for example on empty lines or after the separators. These cases are not formatting issues and should be allowed.

To refine this I would define something like this: `[\w*_:]([ ]{2,}|\\)\n[\w*_:]`
So we require a word (or formatting characters in the first line, followed by *at least* two spaces or a backslash and then a linebreak with a word on the next line.
Please note that I didn't test this so it might need to be adjusted, but this would ensure that this actually creates a line break.

test/validation/test_formatting_validation.py
```diff
@@ -176,6 +177,21 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
     assert actual == expected
 
 
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("This is a line  \nThis is another line", True),
+        ("This is a line\\\nThis is another line", True),
+        ("This is a line\n\nThis is another line", False),
```
TimJentzsch, line 185: I think we will need additional tests for the adjusted regex pattern (see below). For example:

```
("*Word*  \n_Word_", True)  # Line break after formatting characters
("Word    \nWord", True)  # Line break with more than two spaces
("Word\n  \n---  \n\nWord", False)  # Spaces within paragraph line breaks (often happens when transcribing)
("Word \nWord", False)  # Only one space at the end
```

TimJentzsch (changes requested):
Looks good already, I left some comments to improve the regex in order to reduce the number of false positives (very important so that we don't annoy the transcribers).

Also, similar to the other PR, this is missing the response string defined in `tor/strings/en_US.yml`.

TimJentzsch (approved):
This should be good now!

TheLonelyGhost (commented):
I'm a little unclear on the circumstances where we do/don't want this. Is it lines that end with `\` (or 2 spaces) followed by only a single line break? Why is this bad?

Lack of double newline makes sense because markdown. What's the problem with the other format?

TheLonelyGhost:
Nevermind my question now. Necessary context: https://github.com/GrafeasGroup/tor/issues/249. I'll edit the PR description so it autocloses the issue when merged.

tor/validation/formatting_validation.py
```diff
@@ -68,6 +68,21 @@
...
+#
+# This is another line
+# DON'T:
+# Double-spaced line break (note the two spaces at the end of the first line):
+# This is a line
+# This is another line
+# Escaped line break:
+# This is a line\
+# This is another line
+INCORRECT_LINE_BREAK_PATTERN = re.compile(r"[\w*_:]([ ]{2,}|\\)\n[\w*_:]")
```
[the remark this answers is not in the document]
TheLonelyGhost: oof, misunderstood and didn't see the now-linked issue. Ignore me.

PR #263, opened 2021-08-06
Add check for unescaped hashtags

### LINKED ISSUE #254: Add formatting check for hashtags formatted as headings
It often happens that users forget to escape a hashtag, e.g. they do `#Hashtag` instead of `\#Hashtag`. The bot should detect this in the formatting check and instruct the user to escape the number sign. If they intended to make a heading, the user should be instructed to place a space between number sign and heading, e.g. `# Heading` instead of `#Heading`.

TheodoreHua:
Closes #254 

This adds automatic formatting checks for unescaped header markdown, e.g. \#username instead of \\\#username.

### DIFF
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -180,9 +180,9 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
 @pytest.mark.parametrize(
     "test_input,should_match",
     [
-        ("#This is an unescaped header", True),
+        ("#This is an unescaped heading", True),
         ("\#This is properly escaped", False),
-        ("# This is meant to render as a header", False),
+        ("# This is meant to render as a heading", False),
         ("\n     \n   #heading", True),
         ("\n\n\#hashtag1 #hashtag2", False),
         ("\n\n*#hashtag1 #hashtag2*", False),
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -184,8 +184,8 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
         (r"\#This is properly escaped", False),
         ("# This is meant to render as a heading", False),
         ("\n     \n   #heading", True),
-        (r"\n\n\#hashtag1 #hashtag2", False),
-        (r"\n\n*#hashtag1 #hashtag2*", False),
+        ("\n\n\\#hashtag1 #hashtag2", False),
+        ("\n\n*#hashtag1 #hashtag2*", False),
     ]
 )
 def test_check_for_unescaped_heading(test_input: str, should_match: bool) -> None:
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -188,7 +188,7 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
         ("\n\n*#hashtag1 #hashtag2*", False),
     ]
 )
-def test_check_for_unescaped_hashtag(test_input: str, should_match: bool) -> None:
+def test_check_for_unescaped_heading(test_input: str, should_match: bool) -> None:
     """Test if unescaped hashtags are detected."""
     actual = check_for_unescaped_heading(test_input)
     expected = FormattingIssue.UNESCAPED_HEADING if should_match else None
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -14,6 +14,7 @@ from tor.validation.formatting_validation import (
     check_for_heading_with_dashes,
     check_for_malformed_footer,
     check_for_bold_header,
+    check_for_unescaped_hashtag,
     PROPER_SEPARATORS_PATTERN,
     HEADING_WITH_DASHES_PATTERN,
 )
@@ -173,6 +174,21 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
     """Test if fenced code blocks are detected."""
     actual = check_for_fenced_code_block(test_input)
     expected = FormattingIssue.FENCED_CODE_BLOCK if should_match else None
+    assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("#This is an unescaped header", True),
+        ("\#This is properly escaped", False),
+        ("# This is meant to render as a header", False)
+    ]
+)
+def test_check_for_unescaped_hashtag(test_input: str, should_match: bool) -> None:
+    """Test if unescaped hashtags are detected."""
+    actual = check_for_unescaped_hashtag(test_input)
+    expected = FormattingIssue.UNESCAPED_HEADER if should_match else None
     assert actual == expected
 
 
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -182,7 +182,10 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
     [
         ("#This is an unescaped header", True),
         ("\#This is properly escaped", False),
-        ("# This is meant to render as a header", False)
+        ("# This is meant to render as a header", False),
+        ("\n     \n   #heading", True),
+        ("\n\n\#hashtag1 #hashtag2", False),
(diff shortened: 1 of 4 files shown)


test/validation/test_formatting_validation.py
```diff
@@ -176,6 +177,21 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
...
+        ("#This is an unescaped header", True),
+        ("\#This is properly escaped", False),
+        ("# This is meant to render as a header", False)
+    ]
+)
+def test_check_for_unescaped_hashtag(test_input: str, should_match: bool) -> None:
+    """Test if unescaped hashtags are detected."""
+    actual = check_for_unescaped_hashtag(test_input)
+    expected = FormattingIssue.UNESCAPED_HEADER if should_match else None
+    assert actual == expected
```
TimJentzsch, line 192: I would make the check a bit "smarter" in order to reduce the number of false positives (see below).

To do that we need to adjust the test cases as well:
- We'd need to add two leading line breaks to every test case
- I think we can add a few more test cases for common patterns. For example:

```
("\n     \n   #heading", True)  # Spaces on the empty line are allowed, up to three spaces before the heading
("\n\n\#hashtag1 #hashtag2", False)  # Only the first hashtag has to be escaped
("\n\n*#hashtag1 #hashtag2*", False)  # Making it italic also formats it as hashtag
```

tor/strings/en_US.yml
```diff
@@ -259,6 +259,19 @@ formatting_issues:
...
+    \    #Text
+
+
+    would render as:
+
+
+    #Text
+
+
+    To fix this, add a backslash in front of the \\# like so: `\\#Text`. If you meant for it to render as a header, please add a space after the \\# to avoid this warning. i.e. `# Text`."
```
TimJentzsch, line 274: The terminoligy is inconsistent here, the function is named `unescaped_hashtag` and the here it's `unescaped_header`.

I think the text might also be a bit confusing: It says "you have an escaped hashtag", but instead it *needs* to be escaped and is not at the moment.
I would also use the term "heading" instead of "header", to avoid confusion with the header at the top of the transcription.

tor/validation/formatting_issues.py
```diff
@@ -7,3 +7,4 @@ class FormattingIssue(Enum):
     HEADING_WITH_DASHES = "heading_with_dashes"
     MALFORMED_FOOTER = "malformed_footer"
     FENCED_CODE_BLOCK = "fenced_code_block"
+    UNESCAPED_HEADER = "unescaped_header"
```
TimJentzsch, line 10: Inconsistent terminoligy (see above)

tor/validation/formatting_validation.py
```diff
@@ -42,6 +42,12 @@
 # ```
 FENCED_CODE_BLOCK_PATTERN = re.compile("```.*```", re.DOTALL)
 
+# Regex to recognize unescaped hashtags which may render as headers.
+# Example:
+#
+# #Hashtag
+UNESCAPED_HEADER_PATTERN = re.compile(r"([^\\]|^)#[^ ].*")
```
TimJentzsch, line 49: I think we can make this Regex a bit smarter. For example, the bot would currently warn for something like this:

```
\#hashtag1 #hashtag2 #hashtag3
```

or for this:

```
*#hashtag1*
```

However, both examples would display correctly on Reddit.
To fix that, we can require a paragraph linebreak before the number sign. You can take the regex for that from the `PROPER_SEPARATOR_PATTERN`, it should look like this: `\n[ ]*\n[ ]{,3}` (You can have spaces on the empty line and up to three spaces before the heading).

In general, we prefer to have false negatives to having false postitives, the latter could be very annoying for experienced transcribers.

tor/validation/formatting_validation.py
```diff
@@ -119,6 +125,21 @@ def check_for_fenced_code_block(transcription: str) -> Optional[FormattingIssue]
...
+
+    Valid: (backslash)#Text
+    Valid: # Test
+    Invalid: #Test
+    """
+    return (
+        FormattingIssue.UNESCAPED_HEADER
+        if UNESCAPED_HEADER_PATTERN.search(transcription) is not None
+        else None
+    )
```
TimJentzsch, line 140: Here the naming inconsistency again

TimJentzsch (changes requested):
Looking good already! I added a few comments to make the Regex a bit "smarter" and to improve the terminoligy.

test/validation/test_formatting_validation.py
TheodoreHua: Added the testcases and changed the regex a bit, 9fd4c44

tor/validation/formatting_validation.py
TheodoreHua: Changed the Regex to test for the paragraph linebreaks (provided regex) or start of string (this shouldn't happen since there's a header)

TheodoreHua:
That *should* resolve all of the requested changes

tor/strings/en_US.yml
```diff
@@ -259,6 +259,19 @@ formatting_issues:
 
     \    int x = 0;\n
     \   int y = x * 100;"
+  unescaped_heading: "**Unescaped heading syntax**: It looks like you have an unescaped hashtag in your transcription, this may cause it to render as a heading instead of a raw hashtag. For example:
```
TimJentzsch, line 262: Here it still says unescaped hashtag

tor/validation/formatting_validation.py
```diff
@@ -42,6 +42,12 @@
 # ```
 FENCED_CODE_BLOCK_PATTERN = re.compile("```.*```", re.DOTALL)
 
+# Regex to recognize unescaped hashtags which may render as headers.
+# Example:
+#
+# #Hashtag
+UNESCAPED_HEADING_PATTERN = re.compile(r"(\n[ ]*\n[ ]{,3}|^)#[^ ].*")
```
TimJentzsch, line 49: I believe the `.*` at the end can be removed without changing the semantics

TimJentzsch (changes requested):
Just some minor things now, we can probably merge this soon.

tor/validation/formatting_validation.py
TheodoreHua: Yeah, you're right

TheodoreHua:
Also the pytest tests for the escaped headings trigger a deprecation warning stating that `\#` is an invalid escape sequence.

tor/strings/en_US.yml
```diff
@@ -259,6 +259,19 @@ formatting_issues:
 
     \    int x = 0;\n
     \   int y = x * 100;"
+  unescaped_heading: "**Unescaped heading syntax**: It looks like you have an unescaped heading in your transcription, this may cause it to render as a heading instead of a raw hashtag. For example:
```
TimJentzsch, line 262: On second thought I think we should alter this again to make it a bit clearer.

Something like "It looks like you intended to transcribe a hashtag, but made a heading (i.e. "big text") instead. This is caused by not escaping the number sign (`#`) in front of the hashtag. For example: [...]"

I should have noticed that sooner, sorry.

test/validation/test_formatting_validation.py
```diff
@@ -176,6 +177,24 @@ def test_check_for_fenced_code_block(test_input: str, should_match: bool) -> Non
...
 
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("#This is an unescaped heading", True),
+        ("\#This is properly escaped", False),
+        ("# This is meant to render as a heading", False),
+        ("\n     \n   #heading", True),
+        ("\n\n\#hashtag1 #hashtag2", False),
+        ("\n\n*#hashtag1 #hashtag2*", False),
```
TimJentzsch, line 188: You are right, this is indeed an invalid escape sequence. We will have to replace the backslashes in front of the number signs with a double backslash to test this properly.

TimJentzsch (changes requested):
You are right, we need to fix the string escapes for the tests. I also thought of a better way to phrase the error message.

Other than that it looks good!

tor/strings/en_US.yml
TheodoreHua: No problem

TimJentzsch (approved):
Very good! Thanks for your help!

TheodoreHua:
Well, something was completely missed. Actually calling the check function.

PR #266, opened 2021-08-08
Call unescaped heading check

TheodoreHua:
In relation to #263 

The check implemented in that PR wasn't actually called in the `check_formatting_issues` function.

### DIFF
--- a/test/validation/transcriptions/invalid/unescaped-heading.txt
+++ b/test/validation/transcriptions/invalid/unescaped-heading.txt
@@ -1,0 +1,11 @@
+*Image Transcription:*
+
+---
+
+Blah blah
+
+#I hate bugs
+
+---
+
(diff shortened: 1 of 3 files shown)


TimJentzsch (changes requested):
Whoops, let's add a test case here to verify that: https://github.com/GrafeasGroup/tor/blob/main/test/validation/test_formatting_validation.py#L198-L230

test/validation/transcriptions/invalid/unescaped-heading.txt
```diff
@@ -0,0 +1,11 @@
+*Image Transcription:*
+
+---
+
+Blah blah
+
+#I hate bugs
```
TimJentzsch, line 7: Can you add a space between the `#` and the `I hate bugs` here? Otherwise it will also return the unescaped heading formatting issue soon

TimJentzsch (changes requested):
Just one tiny thing, after that we can probably merge

test/validation/transcriptions/invalid/unescaped-heading.txt
TimJentzsch: Wait I'm dumb, I thought I was on a different PR

TimJentzsch (approved):

PR #270, opened 2021-08-12
Fix false positive on multi-heading (i.e. ## Heading)

### LINKED ISSUE #269: Using mutliple `#` for headings is not accepted by the bot
The bots is returning the unescaped heading error when multiple `#` are used in a heading, even when a space is added between them and the text (e.g. `## Heading` will register as "unescaped heading syntax")

TheodoreHua:
Closes #269


itsthejoker (changes requested):
The regex suggestion does fix the issue but opens up a different can of worms with allowing multiple octothorpes in the beginning of a line. See here for an example: https://regex101.com/r/d2gPNh/1

TheodoreHua:
First time I've heard it get called octothorpe, anyway, this should fix both issues https://regex101.com/r/BShoSy/1

itsthejoker:
Yessss thank you!

itsthejoker (approved):

PR #272, opened 2021-08-22
Add formatting check for missing or invalid header

### LINKED ISSUE #271: Add formatting check for missing header
Currently, there is no formatting check to detect if the correct type of header is missing from the transcription. Using this post as an example, and ignoring the other formatting issues in it,
image
we can see that the user has used the title of the post as their header. It should be relatively easy to add a check to see if the transcription contains one of the following: `*Audio Transcription*`, `*Image Transcription*`, or `*Video Transcription*`.

If none of these are included in the transcription, it should be rejected with an explanation of proper header formatting. Whether or not a colon is included and any content after the colon should not be checked, as these can be optional depending on the post.

I would also suggest detecting what type of post (audio/image/video) it is via the title of the submission and specifically only recommending the appropriate header, but not being familiar with the code I don't know how easy to do this would be.

TheodoreHua:
Closes #271 

The code checks if a header of valid type (Image, Video, or Audio) is present in the transcription, using the regex `\*<type>.*\*` (replace `<type>` with valid headers, such as `Image Transcription`) searching above the first separator.

The issue also suggested putting the exact header type the transcription should have in the issue message (e.g. post is an Video, then it should suggest the transcriber put `Video Transcription`).   
I don't really see a good way of doing this as the formatting testing and formatting issue message functions only receive the transcription as a string and not a Reddit or Blossom object (which could be used to get the type from the title or a parameter). I'm more than happy to get proved wrong though.

### DIFF
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -15,6 +15,7 @@ from tor.validation.formatting_validation import (
     check_for_malformed_footer,
     check_for_bold_header,
     check_for_unescaped_heading,
+    check_for_invalid_header,
     PROPER_SEPARATORS_PATTERN,
     HEADING_WITH_DASHES_PATTERN,
 )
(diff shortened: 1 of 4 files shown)


itsthejoker (approved):
This looks great! Thanks!

PR #301, opened 2022-01-14
update slack channel names

itsthejoker:


TimJentzsch (approved):

tor/core/helpers.py
```diff
@@ -292,7 +292,7 @@ def remove_if_required(
         # Selects a message depending on whether the submission is reported or not.
         mod_message = i18n["mod"][f"removed_{'reported' if reported else 'deleted'}"]
         send_to_modchat(
-            mod_message.format(submission.shortlink), cfg, channel="removed_posts"
+            mod_message.format(submission.shortlink), cfg, channel="qa_removed_posts"
```
TimJentzsch, line 295: In hindsight, let's not hard-code these.

Would be awesome if we could use the ID/config system here as well!

tor/core/user_interaction.py
```diff
@@ -69,7 +69,7 @@ def process_coc(
                 f"<{user_url}|u/{username}> has just "
                 f"<{post_url}|accepted the CoC!> {emote}",
                 cfg,
-                channel="new_volunteers",
+                channel="qa_new_volunteers",
```
TimJentzsch, line 72: Here too

TimJentzsch (changes requested):
In hindsight, we probably shouldn't hard-code channel names and use a config-based approach similar to the new Blossom implementation.

I also remembered that we are using `#formatting-issues` in the validation part-- that will have to be changed as well.

PR #308, opened 2022-01-29
Pull the Slack channel IDs from the env variables

### LINKED ISSUE #306: Define Slack channel IDs in `.env` file
Instead of hard-coding the Slack channel IDs, they should be pulled from the environment variables.

Additionally, the formatting channel should use an ID instead of the channel name.

TimJentzsch:
Closes #306.

We currently have the IDs of the Slack channels hard-coded, which is... not great.
In this PR we pull the channel names from the env variables instead (also the formatting channel, which still had the channel name in the previous iteration).

The following variables will have to be set in the `.env` file:
- `SLACK_DEFAULT_CHANNEL_ID`
- `SLACK_COC_ACCEPTED_CHANNEL_ID`
- `SLACK_REMOVED_POST_CHANNEL_ID`
- `SLACK_FORMATTING_ISSUE_CHANNEL_ID`


tor/core/helpers.py
```diff
@@ -13,6 +13,7 @@
 import tor.core
 from tor.core import __version__
 from tor.core.config import config, Config
+from tor.core.config import SLACK_REMOVED_POST_CHANNEL_ID, SLACK_DEFAULT_CHANNEL_ID
```
itsthejoker, line 16: Do we need two import lines for this?

TimJentzsch: Good catch, this happened during the refactoring

itsthejoker (approved):

itsthejoker (approved):

PR #315, opened 2022-02-20
Allow some flexibility in the footer around April Fool's Day

TheLonelyGhost:


tor/validation/formatting_validation.py
```diff
@@ -138,9 +145,14 @@ def check_for_heading_with_dashes(transcription: str) -> Optional[FormattingIssu
 
 def check_for_malformed_footer(transcription: str) -> Optional[FormattingIssue]:
     """Check if the transcription doesn't contain the correct footer."""
+    now = datetime.datetime.now()
+    pattern = FOOTER_PATTERN
+    if now.month == 4 and now.day in range(1, 5):  # some leeway from April 1st
```
itsthejoker, line 150: This should probably be a `datetime.now() in [timedelta range of day March 31st to April 2]`

TheLonelyGhost: I like that a lot better too. Will adjust accordingly.

itsthejoker (dismissed):
approved with the note that I can't test the regex at the moment

TheLonelyGhost:
I don't want to push this out when we could easily just write an automated test for it. Will do that before merging

TimJentzsch:
I briefly validated this with https://regex101.com/ and it seems to work, but yea we should add automated tests for this. Especially that it still accepts the normal footer.

**Feature request**: Allow to use `an` instead of `a` in the footer so that you can adapt it to more words.
Adding a `n?` to the regex should be sufficient.

TheLonelyGhost:
Rebased the branch to account for changes in `main`, but it's the same content as before plus automated tests to check the April Fool's regex.

PR #318, opened 2022-03-13
Disallow characters prior to the header

### LINKED ISSUE #313: Update Invalid Header check to actually verify the header is first
Should just be adding a `^` to the beginning of the regex, but right now anything above the "```" bit, as long as the proper header is actually there, will go through.

image

TheodoreHua:
Fixes #313 

This adds an anchor to check for the beginning of the text in the Invalid Header check, preventing text from coming before the header.

### DIFF
--- a/tor/validation/formatting_validation.py
+++ b/tor/validation/formatting_validation.py
@@ -152,11 +152,12 @@ def check_for_invalid_header(transcription: str) -> Optional[FormattingIssue]:
     Valid: *Video Transcription: Test*
     Valid: *Image Transcription*
     Invalid: *Random Transcription*
+    Invalid: Test *Image Transcription*
     """
     header = transcription.split("---")[0]
     return (
         FormattingIssue.INVALID_HEADER
-        if not any([re.search(r"\*{}.*\*".format(i), header) is not None for i in VALID_HEADERS])
+        if not any([re.search(r"^\**{}.*\*".format(i), header) is not None for i in VALID_HEADERS])
         else None
     )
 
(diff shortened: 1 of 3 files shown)


TheodoreHua:
I'm getting rusty at this

tor/validation/formatting_validation.py
```diff
@@ -246,13 +246,14 @@ def check_for_invalid_header(transcription: str) -> Optional[FormattingIssue]:
...
     Invalid: *Random Transcription*
+    Invalid: Test *Image Transcription*
     """
     header = transcription.split("---")[0]
     return (
         FormattingIssue.INVALID_HEADER
         if not any(
             [
-                re.search(r"\*{}.*\*".format(i), header) is not None
+                re.search(r"^\**{}.*\*".format(i), header) is not None
```
TimJentzsch, line 256: I think we should put a `\s*` after the `^` so that the user can put some whitespace before the header without triggering this check.

I'm also not sure about the `\**` part. That also accepts no asterisk at all or as many as the user wants. I assume this is because of the bold header check, but if anything we should allow an extra one there, not zero or more.

I think in the future I will rewrite the header check completely to give us more control of what we check and which message we display to the user.

TimJentzsch (changes requested):
I left some comments on the regex.

I also think we'll have to adapt the corresponding message in the localization file, indicating that the header is missing OR that it's not at the start of the transcription. That should make it easier for the user to fix the problem (hopefully).

tor/validation/formatting_validation.py
TheodoreHua: Whoops...it was meant to be a `+`

Yeah, the header check is a bit of a mess right now, especially because of the bold header check. Likely better to just merge the two at some point. Either way, resolving this now.

TheodoreHua: This should be resolved as of the latest commit.

TimJentzsch (approved):
Looking good!

PR #319, opened 2022-03-13
Add reference links formatting check

### LINKED ISSUE #286: Add formatting check for reference links
Add an automatic formatting check to detect the accidental usage of reference links.
They have the following format:
```
[<name>]: <link>
```
For example:
```
[lance]: https://en.wikipedia.org/wiki/Holy_Lance "Holy Lance - Wikipedia"
```

Notably, the link doesn't have to be a valid URL. So the following would also be recognized as a reference link:
```
[**Unknown User**]: Oh man i think i just ran out of pain
```

If a line in the transcription has this format, it won't be displayed on New Reddit (and sometimes on Old Reddit).
A user might accidentally use this format, for example when transcribing text messages.

Instead, the first bracket should be escaped:

```
\[**Unknown User**]: Oh man i think i just ran out of pain
```

## References

- [Reddit Markdown Reference](https://www.reddit.com/wiki/markdown#wiki_links)

TheodoreHua:
Closes #286 

This PR adds formatting for reference links, as outlined in the corresponding issue. From my own testing, the reference link formatting doesn't trigger on Reddit when there's text before it (i.e. `Text here [Test]: hello`), however to play it on the safe side considering Reddit's finnicky-ness, the regex I used will still trigger if there's text before. If that is unwanted, replace the regex with `^\[.*]:.*`.

### DIFF
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -353,7 +353,8 @@ def test_check_for_invalid_header(test_input: str, should_match: bool) -> None:
         ("[**Unknown User**]: Oh man i think i just ran out of pain", True),
         ("something something something [**Unknown User**]: Oh man i think i just ran out of pain", True),
         ("[**Unknown User**]:Oh man i think i just ran out of pain", True),
-        (r"\[**Unknown User**]: Oh man i think i just ran out of pain", False)
+        (r"\[**Unknown User**]: Oh man i think i just ran out of pain", False),
+        ("[*The tor bot happily smiles as the entire queue is cleared in CTQ*]", False)
     ]
 )
 def test_check_for_reference_links(test_input: str, should_match: str) -> None:
--- a/test/validation/test_formatting_validation.py
+++ b/test/validation/test_formatting_validation.py
@@ -18,6 +18,7 @@ from tor.validation.formatting_validation import (
     check_for_bold_header,
     check_for_unescaped_heading,
     check_for_invalid_header,
+    check_for_reference_links,
     PROPER_SEPARATORS_PATTERN,
     HEADING_WITH_DASHES_PATTERN,
     UNESCAPED_USERNAME_PATTERN,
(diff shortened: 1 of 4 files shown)


tor/strings/en_US.yml
```diff
@@ -324,10 +324,26 @@ formatting_issues:
   invalid_header: "**Invalid header**: It looks like your header doesn't contain a proper type (Image, Video, or Audio) or is missing italics. Please make sure your post contains a proper header. For example if it's an Image Transcription, it should be the following at the top of your transcription:
 
 
-    \   *Image Transcription: Example*
+    \    *Image Transcription: Example*
```
TimJentzsch, line 327: At first I thought this was intentional, because yaml strings are incredibly frustrating to work with and very unintuitive, but it looks like it was indeed not rendering as a code block. Good catch!

test/validation/test_formatting_validation.py
```diff
@@ -345,6 +346,23 @@ def test_check_for_invalid_header(test_input: str, should_match: bool) -> None:
...
 
+@pytest.mark.parametrize(
+    "test_input,should_match",
+    [
+        ("[lance]: https://en.wikipedia.org/wiki/Holy_Lance \"Holy Lance - Wikipedia\"", True),
+        ("[**Unknown User**]: Oh man i think i just ran out of pain", True),
+        ("something something something [**Unknown User**]: Oh man i think i just ran out of pain", True),
+        ("[**Unknown User**]:Oh man i think i just ran out of pain", True),
+        (r"\[**Unknown User**]: Oh man i think i just ran out of pain", False)
+    ]
```
TimJentzsch, line 357: Let's add a test here that a normal description with `[*Bla bla bla description*]` does not trigger this check.

I think we'll have to be careful that this check doesn't add too many false positives, but if necessary we can also relax it later.

tor/validation/formatting_validation.py
```diff
@@ -75,6 +75,12 @@
 # Image Transcription
 VALID_HEADERS = ["Audio Transcription", "Image Transcription", "Video Transcription"]
 
+# Regex to recognize unescaped reference links.
+# Example:
+#
+# [Test]: Something here
+REFERENCE_LINK_PATTERN = re.compile(r"(?:[^\\]|^)\[.*]:.*")
```
TimJentzsch, line 82: Can you escape the closing `]` too? That makes it a bit easier to read, indicating that it's not a character group.

TimJentzsch (changes requested):
Let's add one more test case, but otherwise this looks solid!

test/validation/test_formatting_validation.py
TheodoreHua: The description test will be added in the soon-to-be commits. I can see how this could cause some false-positives but it could be minimized by only having the regex trigger at the start of lines. From my testing, the reference link formatting only works at the start of the line although I only did very basic testing on desktop old reddit, it could be different elsewhere.

tor/validation/formatting_validation.py
TheodoreHua: Ah, it likely got removed as redundant during a code formatting. Fixed in latest commit.

TimJentzsch (approved):

TimJentzsch:
Nice, thank you!

PR #347, opened 2022-07-24
Deleted CtQ Promo

erenbektas:
Deleted CtQ advertisement parts from bot's responds to successful "claim" and "done"  comments.

### DIFF
--- a/tor/strings/en_US.yml
+++ b/tor/strings/en_US.yml
@@ -79,6 +79,7 @@ responses:
       ✅ The post is yours! Best of luck and thanks for helping!
 
       Does this post break the rules? Please report it! If you don't want this post anymore (or can't complete it), respond to this comment with `unclaim`.
+
       Please respond with `done` when complete so we can check this one off the list!
     already_claimed_by_self: |
       ⚠️ You already claimed this post!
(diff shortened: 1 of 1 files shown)


tor/strings/en_US.yml
```diff
@@ -79,14 +79,6 @@ responses:
       ✅ The post is yours! Best of luck and thanks for helping!
 
       Does this post break the rules? Please report it! If you don't want this post anymore (or can't complete it), respond to this comment with `unclaim`.
-
```
TimJentzsch, line 82: We need to leave an empty line between those so that we get the line break on Reddit

tor/strings/en_US.yml
```diff
@@ -117,14 +109,6 @@ responses:
       If no, please reply with `help` to summon a moderator to resolve the situation.
     completed_transcript: |
       ✅ Awesome, thanks for your help! I'll update your flair to reflect your new count.
-
```
TimJentzsch, line 120: We need to leave an empty line here too

TimJentzsch (changes requested):

TimJentzsch (approved):

PR #353, opened 2022-08-23
Update message when user is blocked

TimJentzsch:
- Renamed `blacklisted` to `blocked`
- Changed response message to
  > ⛔ Sorry you can't transcribe at this time. Please message the mods through modmail.


seeroflights (commented):
thanks for working on this so quickly !!

PR #354, opened 2022-08-24
Update: Response to transcript on Tor post

### LINKED ISSUE #352: Improve response for transcriptions posted on r/ToR
Currently, the bot has a generic response that takes r/nicegirls as example.
Instead, it should look at the submission and extract the actual subreddit and link to the post to make a much more concrete and easier to understand response.

erenbektas:
Closes #352 

The response when a user posts their transcription as a comment on the Tor post now guides the transcribers to the partner post:

---

**OLD:**
      The transcription should be made as a **top-level comment** on the post that you're working on transcribing.
      For example, if the content is on r\/nicegirls, then the transcription should be commented as a reply to that submission on r\/nicegirls.
      
**NEW:**
      The transcription should be made as a **top-level comment** on the \[post on r\/{sub_name}]({post_link}).

### DIFF
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -40,6 +40,11 @@ i18n = translation()
 log = logging.getLogger(__name__)
 
 
+def extract_sub_from_url(url: str) -> str:
+    """returns the sub name from the given url without "r/" at the start."""
+    return url.split("/")[4]
+
+
 @beeline.traced(name="forward_to_slack")
 def forward_to_slack(item: InboxableMixin, cfg: Config) -> None:
     username = str(item.author.name)
@@ -58,9 +63,6 @@ def forward_to_slack(item: InboxableMixin, cfg: Config) -> None:
 
 
 @beeline.traced(name="process_reply")
-def extract_sub_from_url(url: str) -> str:
-    # returns the sub name from the given url without "r/" at the start.
-    return url.split("/")[4]
 def process_reply(reply: Comment, cfg: Config) -> None:
     try:
         log.debug(f"Received reply from {reply.author.name}: {reply.body}")
@@ -70,7 +72,7 @@ def process_reply(reply: Comment, cfg: Config) -> None:
 
         if "image transcription" in r_body or is_comment_transcription(reply, cfg):
             post_link = submission.url
-            sub_name = extract_sub_from_url(submission.url: str) -> str
+            sub_name = extract_sub_from_url(submission.url)
             message = i18n["responses"]["general"]["transcript_on_tor_post"].format(
                 sub_name=sub_name,
                 post_link=post_link,
--- a/tor/core/inbox.py
+++ b/tor/core/inbox.py
@@ -58,6 +58,9 @@ def forward_to_slack(item: InboxableMixin, cfg: Config) -> None:
 
 
 @beeline.traced(name="process_reply")
+def extract_sub_from_url(url: str) -> str:
+    # returns the sub name from the given url without "r/" at the start.
+    return url.split("/")[4]
 def process_reply(reply: Comment, cfg: Config) -> None:
     try:
         log.debug(f"Received reply from {reply.author.name}: {reply.body}")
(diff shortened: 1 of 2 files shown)


tor/core/inbox.py
```diff
@@ -58,6 +58,9 @@ def forward_to_slack(item: InboxableMixin, cfg: Config) -> None:
 
 
 @beeline.traced(name="process_reply")
```
TimJentzsch, line 60: This must be directly above the `def process_reply` line, it's an annotation.

I recommend moving the extract function further up.
A style guideline is also to have two empty lines between function definitions

tor/core/inbox.py
```diff
@@ -58,6 +58,9 @@ def forward_to_slack(item: InboxableMixin, cfg: Config) -> None:
...
+def extract_sub_from_url(url: str) -> str:
+    # returns the sub name from the given url without "r/" at the start.
```
TimJentzsch, line 62: If you document what a function does you should do it as a so-called "doc-string". It's basically just a string as the first item in a function, which has a special meaning.
Then your IDE will also show this string when you hover over the function name.

```suggestion
    """returns the sub name from the given url without "r/" at the start."""
```

tor/core/inbox.py
```diff
@@ -66,7 +69,12 @@ def process_reply(reply: Comment, cfg: Config) -> None:
         r_body = reply.body.lower()  # cache that thing
 
         if "image transcription" in r_body or is_comment_transcription(reply, cfg):
-            message = i18n["responses"]["general"]["transcript_on_tor_post"]
+            post_link = submission.url
+            sub_name = extract_sub_from_url(submission.url: str) -> str
```
TimJentzsch, line 73: The `: str` and `-> str` are type annotations and only valid at the function definition. So `url: str` means that `url` has type string and `-> str` is the same for function return types.
When you use the function, you don't need to provide them, because Python already knows the types from the definition.

```suggestion
            sub_name = extract_sub_from_url(submission.url)
```

The only thing that would be valid here is
```py
sub_name: str = extract_sub_from_url(submission.url)
```
but because we already know that `extract_sub_from_url` returns a `str` from the return type, we don't need to specify the type again.

Also note that in Python type annotations are voluntary, because it's a dynamically typed language. But they can help to detect errors earlier.

TimJentzsch (changes requested):

tor/core/inbox.py
erenbektas: Oh this is nice actually!

tor/core/inbox.py
erenbektas: I see makes sense, thank you!

TimJentzsch (approved):
Looks good to me, let's test it out :)

tor/core/inbox.py
```diff
@@ -66,7 +71,12 @@ def process_reply(reply: Comment, cfg: Config) -> None:
         r_body = reply.body.lower()  # cache that thing
 
         if "image transcription" in r_body or is_comment_transcription(reply, cfg):
-            message = i18n["responses"]["general"]["transcript_on_tor_post"]
+            post_link = submission.url
```
TimJentzsch, line 74:
```suggestion
            post_link = reply.submission.url
```

erenbektas:
Hm, was I suppose to define "submission" somewhere? Isn't that imported from PRAW or something?

tor/core/inbox.py
```diff
@@ -66,7 +71,12 @@ def process_reply(reply: Comment, cfg: Config) -> None:
...
+            sub_name = extract_sub_from_url(submission.url)
```
TimJentzsch, line 75:
```suggestion
            sub_name = extract_sub_from_url(post_link)
```

TimJentzsch (changes requested):
Okay we have to get the `submission` from the `reply` first.

However we then deal with an `url` we get from Reddit, not from Blossom. So we will need to see if the subreddit extraction code still works.

tor/core/inbox.py
erenbektas: Ohh yeah right!! Let's see if it works now

TimJentzsch (approved):

PR #356, opened 2022-08-24
Add missing backslash in unescaped heading string

### LINKED ISSUE #350: Unescaped heading warning does not have backslash in escaped example.
In the unescaped heading warning, the section that explains how to escape the heading doesn't have a backlash in its example:

image

image

The relevant line of code is here:

https://github.com/GrafeasGroup/tor/blob/139a39b76f99b9a6fad4513aef6459ffbbddb8c4/tor/strings/en_US.yml#L334

From my (admittedly minimal) understanding of YAML, I think just adding an escaped backslash like `\\#Text` should be enough.

MurdoMaclachlan:
Closes #350:

> In the unescaped heading warning, the section that explains how to escape the heading doesn't have a backlash in its example. This fixes that.

Hoping I read the contribution guidelines right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [UNRELEASED]
 
 - Track timestamp of last reddit post processed before sending to mod chat (credit: @crhopkins)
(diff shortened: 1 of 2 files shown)


TimJentzsch:
TIL we have contribution guidelines.
FYI the changelog is not actually maintained and might be removed in the future

PR #370, opened 2023-05-17
Make linting consistent with blossom

TimJentzsch:
This PR ports over the linting setup from Blossom to ToR, to have a consistent setup across all our repositories.

- Add `ruff` linter and configuration
- Fix lint errors
- Add pre-commit hook configuration


itsthejoker:
Pytz is deprecated - we'll need to use zoneinfo instead. https://docs.python.org/3/library/zoneinfo.html

itsthejoker:
And I didn't know that `datetime.timezone.utc` was a thing! 🎉

tor/core/users.py
```diff
@@ -17,9 +18,10 @@ class UserDataNotFound(Exception):
 
 
 class User(object):
-    """
+    """A user object.
+
     Usage:
-    from users import User
+    from users import User.
```
itsthejoker, line 24: Not sure why we have a period here?

```suggestion
    from users import User
```

itsthejoker: Honestly it doesn't really matter because this whole section needs to be removed anyway

itsthejoker (approved):

tor/core/users.py
TimJentzsch: Removed the section.

itsthejoker (approved):

